Issue #13, opened 2026-05-02
M1: ryzic music bot foundation


riohno:
Initial milestone of `ryzic`, a self-hostable Discord music bot. Stack: hikari + hikari-lightbulb v3 + lavalink.py + yt-dlp + python-dotenv + aiosqlite. Python 3.13. MIT.

Canonical design docs live at:
- [`docs/plans/M1.md`](https://github.com/VeryLongOrgNameSuchWow/ryzic/blob/main/docs/plans/M1.md) — the plan
- [`docs/plans/M1-review.md`](https://github.com/VeryLongOrgNameSuchWow/ryzic/blob/main/docs/plans/M1-review.md) — pre-implementation correctness/security/UX review
- [`docs/plans/M1-simplify.md`](https://github.com/VeryLongOrgNameSuchWow/ryzic/blob/main/docs/plans/M1-simplify.md) — pre-implementation simplification pass

## Scope (8 PRs per plan §12)

- [x] PR2 — skeleton + /ping bot + tooling (release-please, dependabot, CI) — landed in #1
- [x] PR3a — yt-dlp wrapper + URL validator + tests — landed in #2
- [x] PR3b — audio cache (sqlite-LRU + per-video lock + orphan sweep) — landed in #6
- [x] PR4 — playlist metadata cache (live-first with TTL fallback) — landed in #5
- [x] PR5 — lavalink.py wire-up + voice bridge — landed in #3
- [x] PR7 — Dockerfile + docker-compose + Lavalink config — landed in #4
- [ ] PR6a — /play + ux + voice helper — in flight
- [ ] PR6b — /skip /queue /pause /resume /leave + remove /lltest
- [ ] PR9 — README + SEMVER + manual smoke tests — landed in #7

## Architecture key points

- yt-dlp downloads to local disk; Lavalink plays from local file via `LocalAudioSourceManager`. Cache dir mounted at the same absolute path in both containers (Lavalink read-only).
- Audio cache: sqlite-indexed LRU + size cap, **NO TTL** — survives yt-dlp breakage between patches.
- Playlist metadata cache: live-first, cached fallback on yt-dlp failure.
- Auto-leave on idle: 5min hardcoded.
- Per-guild state: in-memory only; queue dies on restart, audio files survive.
- Self-hostable via docker-compose; configuration via env vars only.

## Process

- Conventional commits → release-please (CHANGELOG.md + version bump + tag + GitHub Release).
- Feature branches + PRs.
[...]

PR #21, opened 2026-05-02
chore(ci): release infrastructure + SHA-pin existing workflows

riohno:
## Summary

Sets up the user-facing release path (Docker image to GHCR on every tag) and aligns workflow tooling with project standards.

## CI changes

- Pin all GitHub Actions to commit SHAs (with version-tag comments) instead of floating tags. Same SHAs as the reference repo where actions overlap. Defends against supply-chain attacks where a maintainer re-tags a pinned major to a poisoned commit.
- Bump `docker/setup-buildx-action` to v4 + `docker/build-push-action` to v7 (dependabot opened PRs for these; doing both here in a single reviewable change instead).
- Adopt the `${{ github.workflow }}-${{ github.ref }}` concurrency group + main-spares-cancellation pattern so concurrent main pushes serialise rather than cancel each other.
- Add `fail-fast: false` for the (single-version) matrix.

## release-please.yml

- Mint an installation token from the `vlonsw-release-please` GitHub App (`actions/create-github-app-token@v2.2.2` SHA-pinned) instead of `GITHUB_TOKEN`. Tags pushed by `GITHUB_TOKEN` do **not** fire downstream workflows (GitHub anti-loop rule); App-token pushes do. Without this, release-please's tag push would silently skip release.yml below.
- Add `workflow_dispatch:` for manual reruns when debugging.
- Pin the release-please-action SHA (v5.0.0).

## NEW release.yml

- Triggers on `push: tags: [v*]` + `workflow_dispatch` (with required `tag` input as escape hatch).
- Builds Docker image with buildx, logs into GHCR via the workflow's `GITHUB_TOKEN` (no extra secret needed), pushes the image with three tags via `docker/metadata-action`: `:X.Y.Z`, `:X.Y`, `:latest`.
- Appends an `image digest` block to the GitHub Release notes so consumers can copy a digest-pinned `docker pull`. Skips silently if no release exists for the tag (manual `git tag` flows).
- Permissions: minimal — `contents: write` (for the release notes edit), `packages: write` (for GHCR push), `id-token: write` (for future attestation hardening).

## NEW release-dry-run.yml

- Runs on PRs that touch packaging-relevant files (`pyproject.toml`, `uv.lock`, `Dockerfile`, `.dockerignore`, `compose.yaml`, `lavalink/application.yml`, the release workflows).
- Builds the image (no push) to catch Dockerfile + compose regressions before they hit a tag push.
- Smoke-tests that the built image fails fast on missing `DISCORD_BOT_TOKEN` — proves the entrypoint is wired (not crashing on import / missing module / wrong Python).

## ⚠️ PRE-MERGE REQUIREMENT

The `vlonsw-release-please` GitHub App's repo secrets must be configured BEFORE this PR merges. Until they are, release-please workflow runs will **fail** on token mint:

- `RELEASE_PLEASE_APP_ID`
- `RELEASE_PLEASE_APP_PRIVATE_KEY`

The App is already installed on this repo. The secrets should be the same values used on `kanbantool-mcp` (App ID is stable across renames; private key is the same).

## Test plan

- [ ] CI on this PR — pending
- [ ] `release-dry-run.yml` fires on this PR (it touches a release workflow file)
- [ ] After merge: next `feat:`-bumping commit on main triggers release-please → release PR opens → merging it tags + fires `release.yml` → image lands at `ghcr.io/VeryLongOrgNameSuchWow/ryzic`

Refs #13.


riohno:
**Blocked on App secrets** — `gh api repos/.../actions/secrets` is empty. This PR converts `release-please.yml` to mint an installation token via `actions/create-github-app-token` (reads `secrets.RELEASE_PLEASE_APP_ID` + `secrets.RELEASE_PLEASE_APP_PRIVATE_KEY`). If we merge before those secrets exist, the next push to `main` will fail.

To unblock, in `Settings → Secrets and variables → Actions → New repository secret`:

- `RELEASE_PLEASE_APP_ID` — numeric App ID from `Settings → Developer settings → GitHub Apps → vlonsw-release-please → About`.
- `RELEASE_PLEASE_APP_PRIVATE_KEY` — full `-----BEGIN/END RSA PRIVATE KEY-----` block from a fresh "Generate a private key" on the same App page (the file downloads as `*.pem`; paste the entire contents).

CI is green, the diff is correct — holding the merge until the secrets land.

riohno (commented):
## /simplify — PR #21 (release infrastructure)

**Verdict:** already tight, with a couple of small wins worth taking

**Estimated trim:** ~10–15 lines (cosmetic), no structural simplification needed

The PR is structurally clean: four workflows with distinct triggers and minimal permissions, SHA-pinning consistent across files, comments mostly explaining WHY (App-token rationale, dry-run gate purpose). KISS/DRY/SRP look healthy. The wins below are cosmetic; none of them are blockers.

### Wins (in order of value)

1. **Drop `type=raw,value=latest` in favour of `flavor` default** — `release.yml:53–56`. `docker/metadata-action` defaults `flavor: latest=auto`, which already attaches `:latest` to non-prerelease semver tags. The explicit `type=raw,value=latest` line is redundant. Drop one line; same behaviour. (-1 line)

2. **Trim the duplicated App-token explanation** — `release.yml:3–8` and `release-please.yml:26–32` both explain the GitHub anti-loop rule and the App-token workaround in roughly the same words. Keep the long form in `release-please.yml` (the workflow that actually mints the token); shrink `release.yml`'s header to one line ("Tag push from release-please's App-token fires this; `GITHUB_TOKEN` tags wouldn't — see `release-please.yml`."). (-5 lines)

3. **Compress the `ref:` rationale on the release checkout** — `release.yml:33–38`. Four-line comment ahead of `ref: ${{ github.event.inputs.tag || github.ref }}` can be one line: `# workflow_dispatch passes a branch ref; pivot to the input tag.` (-3 lines)

4. **`fail-fast: false` is a no-op on a single-version matrix** — `ci.yml:19`. Harmless and forward-looking, but worth a one-line `# kept for when matrix expands` comment so a future reader doesn't think it's load-bearing. (Net 0 lines, +1 line of intent.)

### Items considered, no win

- **Composite `setup-uv-py` action** to dedupe checkout + setup-uv + python install + sync across `lint-and-test` and `integration`. Two call sites only; composite file would cost ~12 lines to save ~10. Not worth it until a third caller appears.
- **Collapse `ci.yml docker-build` into `release-dry-run.yml`.** Different trigger scope (every PR vs packaging-touching PRs only); keeping both is correct.
- **Smoke-test `set +e` / capture / `grep` block** in `release-dry-run.yml`. Could be a one-liner pipe, but the expanded form makes "we expect this to fail" obvious — readability wins.
- **Three temp files in the digest-append step** (`release.yml:78–97`). Pipe-and-`gh release edit --notes "$(...)"` would save 2–3 lines but lose debuggability when the step fails in CI. Keep.
- **Long header comments** on `release-please.yml` and `release.yml`. Both pay rent — they explain non-obvious ordering between the two workflows. Keep.
- **Explicit `permissions:` blocks** — explicitly out of scope per the brief; security best practice.
- **Concurrency group** — same.

Total achievable trim: ~9 lines. Recommend taking #1 and #3 (lowest-risk, smallest-blast-radius); #2 is judgement (redundancy is mild and the duplication aids readability if someone lands on `release.yml` first).

riohno (commented):
## /security-review — PR #21 (release infrastructure)

**Verdict:** ✅ ship as-is (one LOW + a few INFO worth tracking)

**HIGH-severity findings:** 0

I audited the four workflow files for the threat model called out in the brief (App private key handling, App scope, PR-time Docker build trigger, GHCR publish surface, persisted creds, action SHA-pinning, `workflow_dispatch` exposure, tag verification, secret-via-env vs `with:`). Verified all eight SHA pins against the upstream tag refs via the GitHub API — every one is the genuine commit for its claimed `vX.Y.Z` tag. Also confirmed the `vlonsw-release-please` GitHub App is installed at the org with the correct minimal scope (`contents:write`, `metadata:read`, `pull_requests:write`).

### Findings

**[LOW] `release.yml` requests `id-token: write` for unused OIDC** — `.github/workflows/release.yml:26`. The release job currently has no step that consumes an OIDC token (no cosign, no SLSA generator, no AWS/GCP federation). Granting the token preemptively widens the job's blast radius if any future step (or a transitive action update) ever exfiltrates the OIDC JWT — that JWT can be exchanged for cloud creds at any party that trusts your repo. Fix: drop `id-token: write` for now and reintroduce it in the same PR that wires up cosign/SLSA. The "future attestation hardening" comment is fine intent; just gate the permission on actually using it. Defense in depth, not a current exploit.

**[INFO] Anyone with `write` access can manually publish via `workflow_dispatch`** — `.github/workflows/release.yml:16-21`. A compromised maintainer account (or accidental misuse) can publish an arbitrary tag to GHCR `:latest`/`:vX.Y.Z` from the Actions UI, bypassing the release-please flow. The `tag` input is a freeform string. This is an accepted tradeoff per the brief and the workflow comment ("escape hatch when a tag was created some other way") — flagging only so the team is aware of the trust boundary. Mitigations if desired later: tag-protection rules in repo settings, or requiring a manual review on a `release` GitHub Environment.

**[INFO] No tag provenance check before GHCR publish** — `.github/workflows/release.yml:14-15`. A push of any `v*`-shaped tag by anyone with `write` access (or via `workflow_dispatch`) builds + publishes. There is no signature check that the tag was created by the release-please App or signed by a trusted key. Same trust-boundary observation as above; flagged informational because (a) the brief explicitly noted it and (b) requiring signed tags has its own UX cost. If/when the project goes 1.0 + public, consider tag-protection rules + cosign verification of the source commit.

**[INFO] `release.yml` checkout persists `GITHUB_TOKEN` in `.git/config`** — `.github/workflows/release.yml:32`. The default `actions/checkout` persists the workflow `GITHUB_TOKEN` as a `.git/config` extraheader. Any subsequent step in the same job (e.g. a future maintenance task that runs PR-author-influenced code, like a script under `./scripts/`) could read it. Currently the job only runs trusted in-tree steps + pinned actions, so this is theoretical. Belt-and-braces fix: add `persist-credentials: false` to that checkout — neither buildx, build-push, nor `gh release` need a git remote credential.

### Items checked, no finding

- **App private key flow.** Consumed only by `actions/create-github-app-token@v2.2.2` via `with: private-key:` (correct — never via env var, where it'd land in step logs more easily). Minted token is exposed as `steps.app-token.outputs.token` and flows into exactly one downstream step (`release-please-action`) via `with: token:`. Not echoed, not written to `$GITHUB_OUTPUT` by hand, not passed to any third-party action, not exported to env. Action automatically masks its outputs in logs.
- **App scope.** The `vlonsw-release-please` App's actual installed permissions (verified via `gh api`) are `contents:write`, `metadata:read`, `pull_requests:write` — the documented minimum for release-please. No write access to actions, secrets, deployments, packages, or admin endpoints. Even a compromised mint step couldn't escalate beyond release-please's footprint on this repo.
- **`release-dry-run.yml` trigger is `pull_request` (not `pull_request_target`).** Runs against the PR head ref with NO repo secrets and a default `GITHUB_TOKEN` further reduced to `contents: read`. A PR cannot exfiltrate anything because there's nothing to exfiltrate. The Docker build runs in an ephemeral runner with no registry credentials, no GHCR push, no cache other than gha cache (which is partitioned per-ref). Smoke test runs the built image with `--rm` and no mounts; even a malicious `Dockerfile` + `ENTRYPOINT` has nothing in scope to steal.
- **GHCR publish trigger is narrow.** `release.yml` fires only on `push: tags: [v*]` and `workflow_dispatch`. No `workflow_run` chain from `release-dry-run`. PR builds cannot publish to GHCR — no `:pr-N` tag leakage path exists.
- **Tag derivation is GitHub-controlled.** `docker/metadata-action` derives `:vX.Y.Z` / `:vX.Y` / `:latest` from `github.ref` (or the dispatch input). No string interpolation from PR-author-controlled fields into the tag.
- **Action SHA-pinning is complete and correct.** Eight third-party actions across four workflows; all pinned to a 40-char SHA with the version-tag comment matching upstream. Verified each SHA against `api.github.com/repos/{owner}/{repo}/git/ref/tags/{vN}`. No floating `@vN` or `@main` references anywhere.
- **`release-please.yml` has no `actions/checkout`.** Release-please-action handles its own clone with the App token. No risk of leaking the App token via persisted `.git/config` in a follow-up step (and there are no follow-up steps anyway).
- **GHCR login uses the workflow's `GITHUB_TOKEN`** — minimum scope, automatically rotated per run, not the App token. Correct separation.
- **`gh release edit` step uses `GITHUB_TOKEN` via env** — fine, GH CLI requires `GH_TOKEN` env. Output is constrained to release notes (`gh release view`/`gh release edit`); no shell injection vector since `tag` is interpolated from `github.ref`/`inputs.tag`, both controlled by GitHub's tag-validation, not arbitrary user input.
- **Workflow permissions are minimal per job.** `ci.yml` is `contents: read`, `release-dry-run.yml` is `contents: read`, `release-please.yml` is `contents: write` + `pull-requests: write`, `release.yml` is `contents: write` + `packages: write` (+ id-token, see LOW above). Each is the floor for what the job does.

riohno (commented):
## /review — PR #21 (release infrastructure)

**Verdict:** minor revisions

### Findings

- **[MEDIUM] `workflow_dispatch` path silently skips semver tags** — `.github/workflows/release.yml:53-56`. On `workflow_dispatch`, `github.ref` is `refs/heads/main` (the dispatching branch), not a tag. `docker/metadata-action`'s `type=semver,pattern={{version}}` reads the tag from `github.ref` and silently emits no tag when the ref isn't a tag — so the dispatch path will only push `:latest`, skipping `:X.Y.Z` and `:X.Y`. The `actions/checkout` step pivots the working tree but doesn't change `github.ref` for downstream steps. This breaks the documented escape hatch (`gh workflow run release.yml --ref main -f tag=v0.X.Y`).
  Fix: pass the input through explicitly, e.g. `type=semver,pattern={{version}},value=${{ github.event.inputs.tag || github.ref_name }}` (and same for the `{{major}}.{{minor}}` line).

- **[MEDIUM] `:latest` moves backwards on workflow_dispatch for an old tag** — `.github/workflows/release.yml:56`. `type=raw,value=latest` applies unconditionally. Dispatching `release.yml` with an older tag (e.g. patching `v0.1.5` after `v0.5.0` shipped) will overwrite `:latest` with the older image. Realistic during emergency-hotfix flows.
  Fix: gate `latest` on the tag actually being the highest released version, e.g. `type=raw,value=latest,enable=${{ !contains(github.event.inputs.tag || github.ref_name, '-') && github.event_name == 'push' }}` — at minimum, suppress `:latest` on `workflow_dispatch`. The simplest defensible form is `enable=${{ github.event_name == 'push' }}`.

- **[MEDIUM] `release-please-config.json` doesn't enforce SEMVER.md's pre-1.0 bump rules** — `release-please-config.json` (pre-existing, not in PR diff but in scope per the brief). `SEMVER.md` and `RELEASING.md` both state `feat!:` / `BREAKING CHANGE:` should bump MINOR pre-1.0 (`0.X+1.0`). release-please's default for `bump-minor-pre-major` is `false`, which means breaking changes bump MAJOR (`0.5.2` → `1.0.0`). The first `feat!:` after this lands will accidentally cut `v1.0.0`.
  Fix: add `\"bump-minor-pre-major\": true` to the package config. Out-of-band relative to the workflow refactor, but the PR is the right moment to surface it because it's release infrastructure.

- **[LOW] `release.yml` has no concurrency group** — `.github/workflows/release.yml`. Two near-simultaneous tag pushes (or a `push: tags` colliding with a `workflow_dispatch` for the same tag) will run two parallel publishes. Idempotent for the registry but wasteful and noisy.
  Fix: `concurrency: { group: \"release-${{ github.ref }}\", cancel-in-progress: false }` — never cancel in-progress publishes.

- **[LOW] `release-please.yml` and `release-dry-run.yml` lack concurrency groups** — `.github/workflows/release-please.yml`, `.github/workflows/release-dry-run.yml`. Rapid successive pushes to `main` (or PR force-pushes) run parallel jobs. release-please tolerates this but it's noisy; dry-run wastes a builder.
  Fix: standard `concurrency: { group: \"${{ github.workflow }}-${{ github.ref }}\", cancel-in-progress: ${{ github.ref != 'refs/heads/main' }} }` for dry-run; for release-please use `cancel-in-progress: false` so the in-flight release-PR refresh always finishes.

- **[INFO] `<digest unavailable>` fallback publishes a broken pull command** — `.github/workflows/release.yml:93`. If `steps.meta.outputs.digest` is ever empty on a successful push, the appended notes will read `docker pull ghcr.io/...@<digest unavailable>` — silently shipped to consumers. Better to fail explicitly (`if [ -z \"$IMAGE_DIGEST\" ]; then echo \"::error::digest missing\"; exit 1; fi`) and never publish a broken pull command.

- **[INFO] Stale comment references `softprops`** — `.github/workflows/release.yml:69`. The comment says \"softprops just attaches the image-digest reference\" but the implementation uses `gh release edit`. Drop the `softprops` mention to avoid confusing the next reader.

- **[INFO] Build duplication between `ci.yml` docker-build and `release-dry-run.yml`** — A PR that touches `Dockerfile`/`compose.yaml`/`pyproject.toml`/`uv.lock` will run two separate buildx builds in parallel, with separate GHA caches, that produce nearly the same image. The smoke-test step is the unique value of the dry-run job; the build step duplicates ci's. Future option: collapse into a single job that builds once + smoke-tests, or add a job-level `if:` that skips ci's docker-build when dry-run will run. Not worth blocking on.

- **[INFO] `id-token: write` granted but unused** — `.github/workflows/release.yml:26`. Per the PR body it's reserved for future attestation hardening; harmless today (no action requests an OIDC token), but worth a code comment so a future reader doesn't strip it for being unused.

- **[INFO] Pre-merge: repo secrets `RELEASE_PLEASE_APP_ID` + `RELEASE_PLEASE_APP_PRIVATE_KEY` are not yet set** — verified via `gh secret list --repo VeryLongOrgNameSuchWow/ryzic` (empty). PR body already calls this out. Confirming: until those land, the next push to `main` after this merges will fail at `actions/create-github-app-token`, which means no release PR opens until the secrets are in.

### Items checked, no finding

- All 8 SHA pins verified against tag refs via `gh api repos/{owner}/{repo}/git/refs/tags/vX.Y.Z`: actions/checkout v6.0.2, astral-sh/setup-uv v8.1.0, docker/setup-buildx-action v4.0.0, docker/build-push-action v7.1.0, actions/create-github-app-token v2.2.2, googleapis/release-please-action v5.0.0, docker/login-action v4.1.0, docker/metadata-action v6.0.0. All correct.
- `release-please.yml` App-token output (`steps.app-token.outputs.token`) is consumed by the `release-please-action`'s `token:` input. Wired correctly.
- Permissions blocks are minimal and appropriate: `release.yml` (`contents: write`, `packages: write`, `id-token: write`), `release-please.yml` (`contents: write`, `pull-requests: write`), `release-dry-run.yml` (`contents: read`), `ci.yml` (`contents: read`).
- `release-dry-run.yml` correctly uses `pull_request` (not `pull_request_target`); fork PRs run in isolation without secrets exposure. The job uses no secrets, so no fork-trust risk.
- Smoke-test grep `DISCORD_BOT_TOKEN` matches the actual error string raised by `config._require()` in `src/ryzic/config.py:34-37`. Uncaught `ConfigError` will exit non-zero with traceback to stderr, satisfying the smoke test.
- `release-dry-run.yml` paths filter is well-scoped: includes everything that influences the build artifact (`pyproject.toml`, `uv.lock`, `Dockerfile`, `.dockerignore`, `compose.yaml`, `lavalink/application.yml`, both release workflow files); excludes `src/` because import / wiring breakage already fails earlier in `ci.yml`'s lint/test/docker-build jobs.
- `actions/checkout` `ref: ${{ github.event.inputs.tag || github.ref }}` evaluates correctly on both trigger paths (null short-circuits to `github.ref` on `push: tags`).
- `release.yml`'s digest-append step uses `${GITHUB_REF#refs/tags/}` correctly and the `DISPATCH_TAG` env override flips behavior cleanly on `workflow_dispatch`.
- App private key reads only in the single `actions/create-github-app-token` step; not propagated as an output, env, or artifact downstream. Token leak surface is minimal.
- Tag shape contract: `release-please-config.json` has `\"include-v-in-tag\": true`, producing `vX.Y.Z` tags that match `release.yml`'s `tags: [\"v*\"]` trigger. Wired correctly.
- `ci.yml` concurrency change to `cancel-in-progress: ${{ github.ref != 'refs/heads/main' }}` is the right pattern (don't cancel main-branch CI; do cancel PR superseded runs).
- `fail-fast: false` on a 1-version matrix is harmless; behaves the same as without. Forward-compat for adding a 3.14 row.
- PR is clean of secrets / .env / credentials.

PR #23, opened 2026-05-02
feat(commands): /skip /queue /pause /resume /leave + remove /lltest

### LINKED ISSUE #14: PR6b: /skip /queue /pause /resume /leave + remove /lltest
Part of M1 (#13). Last command-surface PR before v0.1.0. Depends on PR6a landing first (provides `voice_check.py` + `ux.py`).

## Scope (plan §12 + §3)

`commands/skip.py`, `queue.py`, `pause.py`, `resume.py`, `leave.py`. Removes the throwaway `/lltest` from PR5.

### `/skip`
- Success: public \`\"Skipped **{escaped_title}**.\"\` + \` \" Queue is empty.\"\` if queue now empty.
- Skip while paused: skip advances; new track stays paused.
- Errors: nothing playing → \`\"Nothing is playing.\"\`

### `/queue`
- No voice-channel requirement.
- Success: embed titled \`\"Queue ({N} tracks · {total_mm:ss})\"\` with **Now playing** field (\`\"[**{title}**]({url})\"\` + progress \`\"{pos_mm:ss} / {len_mm:ss}\"\` + \` \"(paused)\"\` if paused). Description: up to first 10 entries; beyond 10: \`\"… and {N-10} more\"\`.
- Edge: nothing playing + queue empty → ephemeral \`\"Queue is empty and nothing is playing.\"\`
- Edge: playing + queue empty → embed shows only Now playing.

### `/pause`
- Success: public \`\"Paused.\"\`
- Errors: not playing → \`\"Nothing is playing.\"\`; already paused → \`\"Already paused. Use /resume.\"\`

### `/resume`
- Success: public \`\"Resumed.\"\`
- Errors: not playing → \`\"Nothing is playing.\"\`; not paused → \`\"Already playing. Use /pause to pause.\"\`

### `/leave`
- Success: public \`\"Left voice channel. Queue cleared.\"\`
- Errors: bot not in voice → \`\"I'm not in a voice channel.\"\`

### Remove `/lltest`
- Delete `src/ryzic/commands/lltest.py` and the related load in `bot.py`.

## Cross-cutting

All 4 voice-control commands (\`/skip\`, \`/pause\`, \`/resume\`, \`/leave\`) call `ensure_same_voice` from `voice_check.py` (PR6a). \`/queue\` does not.

## Constraints

- Use the exact UX strings above (per plan §3).
- Embeds use \`escape_markdown\` + \`safe_truncate\` from \`ux.py\` (PR6a).
- DM rejected via \`dm_enabled=False\` at command registration.
- Per-PR \`/review\` + \`/security-review\` + \`/simplify\` gates as for prior PRs.

## ~LOC estimate

~400 LOC, 5 command modules + tests.

riohno:
## Summary

Implements the remaining five M1 §3 slash commands and removes the throwaway `/lltest`.

- **`/skip`** — `player.skip()` (never `stop()` then `play()`, per Devoxin/Lavalink.py#153). Skip-while-paused leaves the new track paused. The `" Queue is empty."` suffix lands only when nothing is playing AND the queue is empty after the skip.
- **`/queue`** — read-only embed; no voice precondition. Renders the now-playing track with `mm:ss / mm:ss` progress + `(paused)` suffix; lists the next 10 queued entries; collapses overflow to `… and N more`.
- **`/pause`** / **`/resume`** — `set_pause(True/False)` with the spec'd "already paused / already playing" rejections.
- **`/leave`** — `stop()` + `queue.clear()` + `bot.update_voice_state(guild_id, None)`. The existing `_on_voice_state_update` clears `_voice_ready_events[guild_id]` once Discord confirms the disconnect, so a follow-up `/play` waits on a fresh handshake.

The four playback commands invoke `ensure_same_voice()` first (covered exhaustively in `tests/test_voice_check.py`); `/queue` only checks `ctx.guild_id is not None`. All commands use `contexts=[GUILD]` per the PR6a pattern.

UX strings are spec'd character-for-character per M1 §3. Titles/uploaders flow through `escape_markdown` + `safe_truncate`; durations through `format_duration`.

`/lltest` is gone (`bot.py` `load_extensions` no longer references it; the lavalink_glue comment that referenced it is updated).

## Design notes

- **Stashing TrackInfo on AudioTrack.** Lavalink's `LocalAudioSourceManager` populates `AudioTrack.title`/`uri` from the file path we hand it (`/var/cache/ryzic/audio/.../dQw4w9.audio`), which is unhelpful for embeds. `ux.attach_track_info` / `ux.get_track_info` piggyback the original yt-dlp `TrackInfo` on `AudioTrack.extra` so `/queue` and `/skip` render the original YouTube URL/title/uploader. `/play` attaches the metadata immediately before `player.add()` for every queued track.
- **Requester rendering.** `/queue` shows `(req. by <@id>)` — Discord renders these as username displays in embed bodies and **does not** trigger notifications, so we don't need `GUILD_MEMBERS` intent or a REST round-trip.
- **`/leave` when player.is_connected is False.** Treat as "I'm not in a voice channel." even if Discord's voice-state cache disagrees — the playback POV is the source of truth.

## Test plan

- [x] `uv run ruff check`
- [x] `uv run ruff format --check`
- [x] `uv run ty check`
- [x] `uv run pytest -q` (302 passed, +32 new)
- [ ] Manual smoke per `docs/manual-smoke-tests.md` Core playback / Discord-side rejections sections (deferred to release-time per existing PR pattern; none of the new commands change shipped wire format).

Closes #14


riohno (commented):
## /simplify report — PR #23

Scope: 5 new commands (/skip /queue /pause /resume /leave) + removal of /lltest. PR is +1390/-54 with 32 new tests. Reviewed against the maintainer's "three similar lines is better than a premature abstraction" rule and the project's KISS / DRY / SRP / SOLID guideline. Correctness/security are out of scope (other agents).

Headline: **the PR is mostly tight**. Two small simplifications worth taking; the rest are valid tradeoffs already taken correctly.

---

### Opportunities

#### 1. `attach_track_info`/`get_track_info` is borderline — the "namespace future plugin collisions" justification doesn't apply here (low value, ~10 LOC)

`src/ryzic/ux.py:189-210` wraps three lines:

```python
track.extra[_TRACK_INFO_EXTRA_KEY] = info
# vs.
info = track.extra.get(_TRACK_INFO_EXTRA_KEY)
return info if isinstance(info, TrackInfo) else None
```

Five call sites total: 3 writes in `commands/play.py`, 2 reads in `commands/queue.py` + `commands/skip.py`. The justifying comment ("Namespaced so a future plugin stashing `track_info` cannot collide") cites a hypothetical that doesn't exist in this self-hosted single-app deployment — there are no third-party Lavalink plugins reaching into `AudioTrack.extra` here.

The wrapper does have one genuine value: the `isinstance(info, TrackInfo)` defensive narrow on read. That's worth keeping IF the narrow stays. Net effect of inlining:

- Drop the wrapper pair (~22 LOC including docstrings).
- Promote the constant (`_TRACK_INFO_EXTRA_KEY = "ryzic_track_info"`) to module-level (kept).
- Read sites do `info = track.extra.get(_TRACK_INFO_EXTRA_KEY); if not isinstance(info, TrackInfo): info = None` (~2 LOC each, vs. one helper call).

**Verdict**: Marginal. Keep if you like the symmetry of the named pair, drop if you want one fewer indirection in `ux.py`. Probably not worth churning — the wrapper does pass the "earned helper" test for the read side via the type narrow. **Leaning keep.** Estimated savings if cut: ~15 LOC. Listed for completeness, not pressed.

#### 2. Test fixture duplication: `_track_info` and `_track_with_info` repeated in skip + queue tests (~15 LOC)

Both `tests/test_skip_command.py:33-46` and `tests/test_queue_command.py:34-53` declare local `_track_info()` / `_track_with_info()` helpers with overlapping signatures (queue's is the strict superset — `*, video_id, title, duration_ms, url`; skip's is `title` only). Two definitions, same purpose.

Consolidate by adding `make_track_info()` and `track_with_info()` to `tests/_command_helpers.py`. Both test modules already import from it. This is the same DRY argument that justified the helpers file in the first place — it's **only 2x duplication today**, but if `/seek` or `/now-playing` lands the count climbs.

**Verdict**: **Take it.** Modest LOC saving (~10-15 LOC after dedup), and the helpers file is already the canonical home. Cleaner than the current "almost-shared" state.

#### 3. Test setup duplication for the 4 voice-precondition commands — DON'T parametrize

`tests/test_pause_command.py`, `test_resume_command.py`, `test_leave_command.py`, and `test_skip_command.py` all share these test shapes:

- `test_voice_precondition_short_circuits` — same setup, asserts "I'm not in a voice channel."
- `test_no_lavalink_client_returns_*` — same setup, different message
- `test_no_player_returns_*` — same setup, different message

A parametrized table-test could fold these into one fixture across modules. **Don't.** Each test has subtly different post-conditions (`set_pause_calls == []`, `update_voice_state_calls == []`, `skip_calls == 0`, etc.) — the orthogonal asserts would land in a wide condition table that's harder to read than four short test files. The maintainer's rule applies: "three similar lines is better than a premature abstraction" — these are 5-line tests, not 50-line tests. **Keep as-is.**

#### 4. Comment on the `extra_key` constant overstates the threat (~3 LOC)

`src/ryzic/ux.py:59-62` justifies the `_TRACK_INFO_EXTRA_KEY` constant with "Namespaced so a future plugin stashing `track_info` cannot collide with us." There are no plugins in this single-app deployment. Either drop the constant entirely (inline the literal `"ryzic_track_info"` at 2-3 sites) or shorten the comment to the actual reason (the constant lets you grep all stash points from one place).

**Verdict**: trivial. ~3 LOC trim if the comment is shortened. Not pressed.

---

### Top 3 wins

1. **Consolidate `_track_info`/`_track_with_info` test helpers into `tests/_command_helpers.py`** (~10-15 LOC saved). Same DRY argument that justified the helpers file. Cheap win.
2. **Shorten the `_TRACK_INFO_EXTRA_KEY` justification comment** (~3 LOC saved). The "future plugin collision" framing doesn't fit a single-app deployment.
3. **Optionally inline `attach_track_info`/`get_track_info`** (~15 LOC saved). Borderline — the read-side `isinstance` narrow is the only earned value. Take it only if you find the indirection annoying.

**Total LOC the PR could lose: ~25 (taking #1 + #2) up to ~40 (also taking #3).** That's a ~2-3% trim on a +1390 PR — small in absolute terms, which matches the qualitative read.

---

### Keep as-is (and why)

- **`tests/_command_helpers.py` (183 LOC, shared across 5 test files).** Inlining would mean ~90 LOC of fakes per file × 5 files. The duplication-vs-abstraction tradeoff lands clearly on the abstraction side at 5x. The fakes are deliberately minimal, and the docstring spells out the contract. Real DRY win.

- **Five separate command files (`skip.py` / `queue.py` / `pause.py` / `resume.py` / `leave.py`).** CLAUDE.md project conventions explicitly say "one slash command per file (lightbulb extension pattern)." Each file is 50-74 LOC and stands alone. Lightbulb's `Loader` does support multiple commands per module, so the choice is convention-driven, not technical. Keep — convention beats marginal LOC.

- **`build_queue_embed` in `ux.py` (one consumer, kwargs-heavy).** The other embed builders (`build_queued_track_embed`, `build_queued_playlist_embed` from PR6a) live in `ux.py` and are also single-consumer. Convention-driven again. The kwargs argument is `now_playing`, `now_playing_position_ms`, `paused`, `queue` — all genuinely different inputs, no over-parameterization. Keep.

- **`/queue` first-10 + overflow logic (`ux.py:250-269`, ~20 LOC).** Already minimal: slice → list-comp → if-overflow-append → join. Can't credibly shrink without losing readability. Keep.

- **`tool.ty.environment.root = ["./src", "./"]` widening.** The `./` widening exists exclusively so ty can resolve `from tests._command_helpers import ...`. The alternatives (move helpers into `conftest.py` — but they're classes, not fixtures; move into `src/ryzic/_test_helpers.py` — pollutes shipped package; relative imports — fragile across the test runner) are all worse. The 4-line comment justifies the widening clearly. Keep.

- **Comments narrating WHAT.** None found in the new code. All comments explain WHY (the `DefaultPlayer.skip` semantics doc in `skip.py:66-71`, the `stop()`-then-`clear()` race rationale in `leave.py:53-55`, the defensive-empty branch in `queue.py:53-56`, etc.). Comment quality is consistent with the rest of the codebase.

---

### Summary

**This PR is already tight.** The maintainer's "three similar lines beats a premature abstraction" rule was respected at every fork point I inspected: each command file standing alone, the `build_queue_embed` builder for one consumer, the 4 voice-precondition test shapes left as-is across modules. The `_command_helpers.py` extraction was the right call (5x duplication beats 90 LOC × 5 inline fakes).

The two minor wins above are worth taking IF the implementer is touching the file anyway. None of them block merge. If the choice is "trim now" vs. "merge as-is and trim in a follow-up," the simplification cost is low enough either way.

riohno (commented):
## /security-review report — PR #23

### Severity table

| ID | Severity | Area | Title |
|----|----------|------|-------|
| M-1 | MEDIUM | `/leave` + `/skip` semantics, audio_cache lifecycle | `/leave` clears the queue without releasing audio-cache pins for queued tracks |
| L-1 | LOW | embed/Markdown formatting | `TrackInfo.url` interpolated into `[…](url)` without escaping `)` / `\` |
| L-2 | LOW | embed display | Hostile yt-dlp title containing `<@id>` / `<@&id>` / `<#id>` literal text can produce a username/role/channel-display visual spoof in `/queue` (no notification) |
| L-3 | LOW | tests / fakes | `tests/_command_helpers.FakePlayer` mimics `DefaultPlayer.{skip,stop}` synchronously and skips the audio-cache release contract — hides M-1-class bugs from unit tests |

No HIGH or CRITICAL findings.

---

### M-1 (MEDIUM) — `/leave` leaks audio-cache pins for queued tracks

**Where:** `src/ryzic/commands/leave.py:99-103`

```python
await player.stop()
player.queue.clear()

bot = cast(hikari.GatewayBot, ctx.client.app)
await bot.update_voice_state(guild_id, None)
```

**Threat model:** `audio_cache.AudioCache.get_or_download` increments `self._in_use[video_id]` for every successful download (every `_load_one` call in `/play`). The release contract (`audio_cache.py:224-239` docstring) requires exactly one `release` per `get_or_download`, normally driven by `EventHandler.on_track_end → _release_track` which fires when Lavalink dispatches `TrackEndEvent` for the *playing* track.

`player.queue.clear()` discards queued (never-played) `AudioTrack` objects in pure Python. Lavalink never sees them, so it never emits `TrackEndEvent` for them. The pin counts in `audio_cache._in_use` for those `video_id`s are never decremented. Every leaked pin permanently disables LRU eviction for that file (`_evict_to_fit` skips `_in_use.get(video_id, 0) > 0`).

**Impact:** Bot operator DoS — slow disk-fill via repeated `/play <large playlist> /leave` cycles. Cache `max_bytes` cap becomes ineffective for any leaked entry. State resets only on bot restart.

**Pre-existing exposure:** `EventHandler.on_websocket_closed` (code 4014) and `on_node_disconnected` use the same bare `queue.clear()` and have the same leak. `/leave` adds a *user-triggerable*, no-precondition path on top of pre-existing event-driven paths.

**Recommended fix:** Iterate the queue and release each track before clearing (parallels what `_release_track` does for the current track via `TrackEndEvent`). E.g.:

```python
await player.stop()
for queued in player.queue:
    await lavalink_glue._release_track(queued)
player.queue.clear()
```

(Either expose `_release_track` as `release_track` or factor the cache lookup into a tiny shared helper.) Apply the same fix to `on_websocket_closed` and `on_node_disconnected` while you're there.

---

### L-1 (LOW) — `TrackInfo.url` interpolated into `[label](url)` without escaping

**Where:** `src/ryzic/ux.py:239` (`build_queue_embed` — Now-playing field) and `src/ryzic/ux.py:261` (`_build_queue_description` — queued entries). Also `src/ryzic/ux.py:119` in the pre-existing `build_queued_track_embed`.

```python
f"[**{np_title}**]({now_playing.url})\n{progress}"
f"{idx}. [{escape_markdown(info.title)}]({info.url}) — …"
```

**Threat model:** `escape_markdown` is applied to titles but **not** to URLs. Markdown link syntax `[label](url)` is terminated by the first unescaped `)`. yt-dlp's `webpage_url`/`url` fields are surfaced raw to `_track_from_info` and `_entry_from_flat` (`src/ryzic/ytdlp.py:155`, `:175`) with no validation beyond truthiness.

**Realistic exposure:** YouTube's own extractor produces canonical clean URLs (`https://www.youtube.com/watch?v=…`), so a malicious URL would require either (a) yt-dlp regression, (b) a future allow-list expansion to a domain that returns dirty URLs, or (c) `_entry_from_flat`'s entry-level URL falling through to a hostile-supplied value. The `_VIDEO_ID_RE` validation contains the youtu.be fallback path, but the `entry.get("url")` first-choice path is unvalidated.

**Recommended fix (defense in depth):** Either validate the URL via `is_supported_url` before stashing it, or escape `)` and `\` in the URL before interpolating. The cheapest patch is to drop the URL on a failing `is_supported_url` check inside `_entry_from_flat` / `_track_from_info` and substitute the youtu.be canonical fallback.

---

### L-2 (LOW) — `<@id>` / `<@&id>` / `<#id>` literals in titles produce visual spoofs

**Where:** `src/ryzic/ux.py:261`, the description line interpolating `escape_markdown(info.title)`.

**Threat model:** `escape_markdown` covers `\ [ ] ( ) * _ ~ \` | >` (`src/ryzic/ux.py:49`). It does **not** escape `<`, `@`, `&`, or `#`. A track titled `"<@123456789012345>"` will render in the embed body as the user's display name (Discord client-side substitution).

**Notification risk:** None. hikari's `_build_message_payload` always emits `allowed_mentions` (`hikari/impl/rest.py:1549-1552`) defaulting to `{parse: []}` (`hikari/internal/mentions.py:67-90`), which suppresses pings for `<@id>` / `<@&id>` / `@everyone` / `@here`. Embed body mentions also do not notify per Discord behavior. Two layers of protection.

**Residual risk:** Visual deception only. A hostile uploader can craft a title that *looks like* "@admin queued this track" inside `/queue`. No escalation, no ping, no token leak.

**Recommended fix (defense in depth):** Either (a) extend `_MARKDOWN_CHARS` with `<` (and `@`?) so `<@id>` becomes `\<@id>` literal text, or (b) replace `<` with a similar look-alike before interpolation. The README/embed already trusts yt-dlp output for titles; this is genuinely a polish/UX item, not a security blocker.

---

### L-3 (LOW) — `tests/_command_helpers.FakePlayer` doesn't model the audio-cache release contract

**Where:** `tests/_command_helpers.py:127-138`

```python
async def skip(self) -> None:
    self.skip_calls += 1
    if self.queue:
        self.current = self.queue.pop(0)
    else:
        self.current = None

async def stop(self) -> None:
    self.stop_calls += 1
    self.current = None
```

**Threat model:** Real `lavalink.DefaultPlayer.skip` / `.stop` cause the Lavalink server to emit `TrackEndEvent`, which `EventHandler.on_track_end` handles by calling `audio_cache._release_track`. The fakes pop synchronously and dispatch nothing. Tests therefore cannot detect missing release-pin handling — which is exactly what enables M-1 to slip through `tests/test_leave_command.py`.

**Other minor drift:**
- `FakePlayer.set_pause` updates `self.paused` synchronously; real `DefaultPlayer.set_pause` only awaits `node.update_player(...)` and the local `paused` field is updated when the gateway echoes back. Doesn't hide bugs in the current command set, but is worth knowing.
- `FakePlayer.skip` always pops `[0]`; real `DefaultPlayer.play` picks `randrange(queue)` when `shuffle=True`. Ryzic doesn't expose shuffle yet; non-issue.

**Recommended fix:** Either model `_in_use` in `FakePlayer.{skip, stop}` (post-skip release) or — simpler — add a `tests/test_leave_command.py` case that asserts the audio-cache release is invoked for every queued track by patching `audio_cache._in_use` directly and counting before/after. This test would currently fail and pin down M-1.

---

### Items checked, no finding

1. **`@everyone` / `<@&role_id>` injection in `/queue` titles or content.** hikari's `_build_message_payload` always emits `allowed_mentions={parse: []}` for ctx.respond and `bot.rest.create_message` (verified at `hikari/impl/rest.py:1549-1552`, `hikari/internal/mentions.py:67-90`). Two-layer protection (Discord embed-body mentions also don't notify). **PASS.**
2. **`<@id>` requester spoofing.** Sourced from `ctx.user.id` exclusively (`commands/play.py:240, 314, 324`); no caller-controlled override path. `(req. by <@{requester_id}>)` in `_build_queue_description` casts via `int()` to drop any non-numeric junk. **PASS.**
3. **`AudioTrack.extra` injection.** `attach_track_info` writes a typed `TrackInfo` dataclass under a namespaced key (`"ryzic_track_info"`); `get_track_info` validates `isinstance(info, TrackInfo)` before returning. yt-dlp output cannot smuggle a non-`TrackInfo` past the type check. **PASS.**
4. **`/skip` cache release.** `DefaultPlayer.skip` → `play()` → either `play_track(next)` (server emits `TrackEndEvent` for the previous track, reason REPLACED → `_release_track`) or `stop()` + `QueueEndEvent` (server emits `TrackEndEvent`, reason STOPPED → `_release_track`). Verified against `lavalink/player.py:223-340`. The currently-playing track's pin is cycled correctly via the event handler. **PASS** (independently of M-1, which is about *queued* tracks during `/leave`).
5. **`tool.ty.environment.root = ["./src", "./"]` widening.** `ty` only reads `.py` files; `.env.example` contains template defaults (no real secrets, password literal `youshallnotpass` is the public Lavalink default), `.gitignore` covers real secrets, the `lavalink/` repo dir is YAML-only, `tests/` is mock-only with no embedded credentials. The widening's only side effect is letting cross-test imports (`from tests._command_helpers import …`) resolve. **PASS.**
6. **Tests reach external services.** All five new test files mock at the lavalink_glue/audio_cache/yt-dlp boundaries via fakes; no `httpretty`, `aiohttp`, real `socket`, or `YoutubeDL` instantiation in the new tests. Integration tests (`tests/integration/test_lavalink_smoke.py`) gated by `pytest.mark.integration`, skipped by default `addopts = "-m 'not integration'"`. **PASS.**
7. **`/lltest` removal.** Pure deletion + `bot.py` `load_extensions` updated. The lavalink-node `name="ryzic-default"` keeping `host:port` out of node identifiers is preserved (`lavalink_glue.py:493`); the comment update reflects the new rationale (log lines instead of `/lltest`). **PASS.**

---

### Verdict

**fixes recommended** (not merge-blocking).

M-1 is a real correctness/DoS bug but consistent with pre-existing `queue.clear()` paths in the same module, and recoverable on bot restart. Worth fixing in a small follow-up PR (cache-release helper, applied to `/leave`, `on_websocket_closed`, `on_node_disconnected`). L-1, L-2, L-3 are defense-in-depth polish, deferable to dedicated cleanup PRs.

The PR's deliberate boundary choices (escape titles, leave URLs raw, lean on hikari's defaults for mentions) are all defensible and the tests are well-scoped — the M-1 gap is genuinely hidden by the test fakes, not by sloppy review.

riohno (commented):
## /review report — PR #23

Branch `feat/remaining-commands` adds `/skip /queue /pause /resume /leave` and removes the throwaway `/lltest`. 1390/-54 LOC, 302 tests passing locally. Reviewed against `docs/plans/M1.md` §3 + cross-cutting rules in §3 prelude + `CONTRIBUTING.md` standards.

### Severity table

| Sev | Area | Finding |
|---|---|---|
| MEDIUM | `/skip` | `DefaultPlayer.skip()` does not preserve `paused` server-side — the spec's "skip-while-paused leaves new track paused" is not actually implemented. The PR's own docstring asserts the opposite without test coverage. |
| LOW | `/queue` title | `"Queue (1 tracks · 0:30)"` — spec literal has the same wart, but implementer is faithful to the literal so flagging for awareness only. |
| LOW | `/queue` empty edge | When playing + queue empty, the embed still renders `Queue (0 tracks · 0:00)` as title. Plan §3 line 138 says "embed shows only Now playing" — defensible reading either way. |
| LOW | `/leave` "not in voice" semantics | Sends "I'm not in a voice channel." even when Discord cache disagrees with `player.is_connected`. Implementer documents this as deliberate; minor UX wart for the post-restart edge case. |
| LOW | `/skip` fallback path | When `attach_track_info` was bypassed, the fallback uses `player.current.title` which under `LocalAudioSourceManager` is the file path — user would see `Skipped **/var/cache/ryzic/audio/dQ/dQw4w9.audio**.`. Defensive-only path; production callers always attach. |
| LOW | `tool.ty.environment.root` widening | `["./src", "./"]` exposes top-level `tests/` to ty. Other top-level dirs (`docs/`, `lavalink/`, etc.) are not Python packages so collision risk is nil. Pragmatic; cleaner alternative would be moving helpers under `tests/conftest.py` but that's a fixture/import shape change. |
| INFO | `extra` dict reuse | `requester` and `ryzic_track_info` coexist in `AudioTrack.extra`. Already the lavalink.py-canonical pattern (their own `requester` lives in `extra`). Namespacing key `"ryzic_track_info"` is good defensive practice. |

### 1. Plan §3 UX strings

All five commands match plan literals character-for-character, verified against `M1.md` lines 128–146:

- `/skip` success: `Skipped **{escaped_title}**.` plus `" Queue is empty."` suffix — match.
- `/skip` nothing playing: `Nothing is playing.` — match.
- `/queue` empty edge: `Queue is empty and nothing is playing.` (ephemeral) — match.
- `/queue` title: `Queue ({N} tracks · {total_mm:ss})` — match (literal, including pluralization wart).
- `/queue` Now playing field: `[**{title}**]({url})\n{pos_mm:ss} / {len_mm:ss}` plus ` (paused)` suffix — match.
- `/queue` body lines: `{i}. [{title}]({url}) — {mm:ss} (req. by <@{id}>)` — match.
- `/queue` overflow: `… and {N-10} more` — match.
- `/pause` success/errors: `Paused.` / `Nothing is playing.` / `Already paused. Use /resume.` — match.
- `/resume` success/errors: `Resumed.` / `Nothing is playing.` / `Already playing. Use /pause to pause.` — match.
- `/leave` success/error: `Left voice channel. Queue cleared.` / `I'm not in a voice channel.` — match.

### 2. `/queue` embed shape

`build_queue_embed` (`src/ryzic/ux.py:204-234`) renders:

- Title: `f"Queue ({queue_count} tracks · {format_duration(queue_total_ms)})"` — `queue_count` excludes now-playing, faithful to plan literal `N tracks`. Spec doesn't disambiguate; implementer's reading (queued-only) is the natural one given the embed has a separate "Now playing" field.
- Now playing field: `[**title**](url)\n{pos} / {len}` plus ` (paused)` if paused — matches spec.
- Description body: enumerated 1..10 list + `… and N more` overflow line — matches spec.
- Edge cases:
  - DM (no guild_id) → ephemeral `"Run /queue in a server."` — defense-in-depth beyond `contexts=[GUILD]`.
  - Nothing playing + queue empty → ephemeral `"Queue is empty and nothing is playing."` — match.
  - Playing + queue empty → embed with title `Queue (0 tracks · 0:00)` and only Now playing field. Spec says "shows only Now playing"; reading the title as descriptive metadata is reasonable, but a stricter reading would suppress the "0 tracks" header when no queue exists. Cosmetic ambiguity.

Minor: `_QUEUE_PREVIEW_MAX = 10` is a private constant in `ux.py`. Plan §3 hardcodes 10, so this is fine; no need to expose as config.

### 3. Cross-cutting voice check

`ensure_same_voice` is invoked first on `/skip`, `/pause`, `/resume`, `/leave`. NOT invoked on `/queue` (read-only, per spec). All 5 commands share the `contexts=[hikari.ApplicationContextType.GUILD]` registration. `voice_check.ensure_same_voice` (PR6a, unchanged here) does the cache-only `get_voice_state` lookup. Test `tests/test_voice_check.py` already covers branches; PR6b spot-checks wiring on each command (`test_voice_precondition_short_circuits`).

### 4. `/lltest` removal

- `src/ryzic/commands/lltest.py` deleted: confirmed.
- `bot.py` `load_extensions` no longer references it: confirmed (also re-orders `/play` to first which is cosmetic but fine).
- `lavalink_glue.py` updated comment that previously said `/lltest surfaces node.name to invokers` now says `log lines (which include node.name) don't leak the address` — confirmed at line 491. No remaining `lltest` references in the PR branch (`git ls-tree | grep lltest` returns nothing).
- No leftover test file (the PR5 work didn't add a `test_lltest.py`).

### 5. `attach_track_info` / `get_track_info` design

`AudioTrack.extra` is the **canonical** lavalink.py mechanism for user-attached data. The library itself stores `requester` in `extra['requester']` (verified in lavalink.py source: `__slots__` includes `extra`, the class docstring documents `extra: Dict[str, Any] — Any extra properties given to this AudioTrack will be stored here`). Using `extra['ryzic_track_info']` is idiomatic, namespaced (collision-safe), and the alternative — a parallel `dict[guild_id+identifier, TrackInfo]` keyed by player+track — would be strictly worse (cleanup hooks on track removal, lookup overhead, harder mocking).

The two helper functions are thin (3 lines each), live in `ux.py` with the embed builders that consume them, and are tested transitively via the queue/skip command tests. Clean SRP: `ux.py` owns "how a track surfaces to users", track metadata stash is part of that.

One nit: the `isinstance(info, TrackInfo)` guard in `get_track_info` is good defense; it would also catch a future bug where a different module wrote a string under the same key.

### 6. Spec interpretation calls

**`/skip` "Queue is empty." trigger:** Implementer appends iff `player.current is None AND not player.queue` after the skip. Plan §3 line 129 reads literally `+ " Queue is empty." if queue now empty`. The implementer's reading is more correct than the literal — if `skip()` promotes the next track and `queue` becomes empty, saying "queue is empty" while the just-promoted track plays is misleading. The two-clause check (`current is None AND queue empty`) is the right interpretation of intent. Defensible, well-documented in the inline comment at `commands/skip.py:69-74`.

**`/leave` when not connected:** Implementer treats `player is None or not player.is_connected` as "not in voice channel." Plan §3 line 146 only says `bot not in voice → "I'm not in a voice channel."` — doesn't disambiguate Discord cache vs. lavalink player POV. Implementer's choice to use playback POV as source of truth is sensible: if `player.is_connected` is False, a `/play` would need to re-handshake anyway. The post-restart edge case (Discord shows bot in voice but lavalink lost connection) gives the user the actionable message they need.

**`/skip` while paused — paused-preservation issue:** This is the one I want to flag for follow-up. The spec at M1 §3 line 130 says: *"Skip while paused: skip advances; new track stays paused; user must `/resume`."* The implementer's docstring claims `DefaultPlayer.skip` calls `play()` which "intentionally leaves `paused` alone." This is **partially incorrect** at the lavalink.py layer:

- `DefaultPlayer.skip()` is literally `await self.play()` — no args.
- `DefaultPlayer.play(pause=MISSING)` does not pass `paused=` to `play_track`.
- `BasePlayer.play_track` only sets `options['paused']` if `pause is not MISSING`.
- `Node.update_player(paused=MISSING)` does not include `paused` in the JSON body.

Server-side behavior with no `paused` field on a track-replace is **undocumented in lavalink.py source**. Lavalink REST docs imply unspecified fields are not modified, but a track-replace might reset paused state. Whether the spec's "stays paused" actually holds depends on Lavalink server's implementation, which can't be verified from the lib alone — needs a manual smoke test (`/play` → `/pause` → `/skip` and observe whether the new track auto-plays or stays paused).

**Recommendation:** Either (a) explicit `await player.set_pause(True)` after `player.skip()` if `player.paused` was True before (defensive), or (b) call out in the manual smoke test list and in the docstring that the spec's "stays paused" depends on Lavalink server behavior, not the client. The current code does neither — the docstring confidently asserts behavior that the client can't guarantee. Test `tests/test_skip_command.py` does not assert paused-preservation across skip.

### 7. `/pause` / `/resume` while not in expected state

- `/pause` when nothing playing: ephemeral `"Nothing is playing."` — match.
- `/pause` when already paused: ephemeral `"Already paused. Use /resume."` — match.
- `/resume` when nothing playing: ephemeral `"Nothing is playing."` — match.
- `/resume` when not paused: ephemeral `"Already playing. Use /pause to pause."` — match.

`is_playing = is_connected AND current is not None` (independent of `paused`), so the branch order (`is_playing` → `paused`) is correct: paused-but-current-set passes the `is_playing` gate and reaches the `paused` branch.

### 8. DM rejection consistency

All 5 commands register `contexts=[hikari.ApplicationContextType.GUILD]` (lightbulb v3 / Discord-API equivalent of v2 `dm_enabled=False`). Runtime defense:

- `/skip /pause /resume /leave`: `ensure_same_voice` does `if guild_id is None: respond "Run this in a server." (ephemeral); return None`.
- `/queue`: explicit `if guild_id is None: respond "Run /queue in a server." (ephemeral); return`.

Two slightly different DM-rejection strings ("Run this in a server." vs "Run /queue in a server."). Defensible because `/queue` is the only one that bypasses `ensure_same_voice`, but a small style nit — could share one string. Not worth refactoring.

### 9. `tool.ty.environment.root` widening

`["./src", "./"]` adds the project root to ty's import resolution so `from tests._command_helpers import ...` resolves under the same path pytest uses. Side effects:

- `tests/` is now type-checked as a package by ty (was already, via `./src` not exposing `tests` it relied on pytest's path injection — now ty sees it natively).
- Other top-level dirs (`docs/`, `lavalink/`, `.github/`) are not Python packages, so no collision.
- `ruff check`, `ruff format --check`, `ty check`, `pytest -q` all green on the PR branch (verified locally).

Cleaner alternative: put the helpers in `tests/conftest.py` (auto-discovered, no import path needed) or use a `pytest_plugins` shim. Implementer chose the explicit-import approach which is more readable in test files. Minor pragmatic tradeoff; the inline pyproject.toml comment explains the rationale clearly.

### 10. KISS / DRY / SRP / SOLID + comments policy

**KISS:** All five new command modules follow the same minimal shape (`Loader` + `SlashCommand` + `_handle_*` async helper). No premature abstractions. Each module is ~50–75 LOC. Good.

**DRY:** Shared test fakes in `tests/_command_helpers.py` are the right call — saves ~5× duplication of the slim hikari/lavalink stand-ins. The `_handle_*` extraction (vs putting logic inside the `invoke` method) lets tests bypass lightbulb's invocation machinery — same pattern as PR6a's `play.py:_handle_play`. Consistent.

A near-DRY-violation: the `if ll_client is None / if player is None or not player.is_playing: respond "Nothing is playing."` block is repeated in `/pause /resume /skip` (3×). Could be factored into a `_get_active_player(ctx) -> DefaultPlayer | None` helper that responds + returns None on failure. But the duplication is only 8 lines × 3 — extraction would add a layer of indirection for marginal payoff. Acceptable per KISS.

**SRP:** `ux.attach_track_info`/`get_track_info` belong in `ux.py` (track-metadata-for-display lives next to embed builders that consume it). Each command module owns one command. Voice precondition stays in `voice_check.py`. Clean separation.

**SOLID:** No interface inversions warranted at this size. The `Loader` + decorator pattern is lightbulb's idiom, not invented here. Module-level singletons (`get_lavalink_client`) — already established in PR5 and tested via `_set_lavalink_client_for_test`.

**Comments policy** (CONTRIBUTING: "no comments narrating WHAT — only WHY when non-obvious"):

- `commands/skip.py` module docstring: explains WHY `skip()` over `stop()+play()` (Lavalink.py#153) — good.
- `commands/skip.py:69-74` inline comment on the "Queue is empty" suffix logic: explains the WHY (avoid misleading "empty" while a just-promoted track plays) — good.
- `commands/leave.py:76-79` inline comment on `stop` before `update_voice_state`: explains WHY (auto-advance race) — good.
- `commands/queue.py:54-58` defensive comment on missing-metadata path: explains WHY (future code path that bypasses attach) — good.
- `ux.py:206-211` `attach_track_info` docstring: explains WHY (local source manager populates URI from file path, useless for embeds) — good.
- `pyproject.toml:62-65` comment on `tool.ty.environment.root` widening: good rationale.

No narration-of-WHAT comments spotted. All inline comments justify a non-obvious choice.

### Verdict

**Minor revisions.** The implementation is otherwise FOSS-grade and faithfully renders M1 §3. The single substantive issue is the **`/skip` paused-preservation claim**: the docstring asserts behavior that lavalink.py's `skip()` cannot actually guarantee from the client side. Either:

1. Add `if was_paused: await player.set_pause(True)` after `player.skip()` in `commands/skip.py:_handle_skip`, with a test that verifies pause survives a skip, OR
2. Rewrite the docstring + add a manual-smoke entry: "Verify /play → /pause → /skip leaves the new track paused (relies on Lavalink server's behavior with track-replace + no paused field)."

Either is acceptable; option (1) is defensive and removes ambiguity. Everything else (queue title pluralization, description-when-empty, /leave fallback semantics) is plan-faithful or defensible-implementer-call territory.

The `/lltest` removal is clean. The `attach_track_info` design is the right call — uses lavalink.py's own canonical extension mechanism. Test coverage is thorough (32 new tests against the five new commands + the queue embed builder). The shared `_command_helpers.py` is well-scoped and well-documented. `tool.ty.environment.root` widening is pragmatic and consequence-free.

PR #28, opened 2026-05-02
fix(audio-cache): release pins for queued tracks on queue-clear paths

### LINKED ISSUE #24: Audio cache: pin leak when queue is cleared without playing each track
## Summary

`audio_cache` pins are released by `EventHandler._release_track` on `TrackEndEvent` (reasons FINISHED / REPLACED / STOPPED / LOAD_FAILED). Code paths that **clear the queue without ever playing a queued track** leak that track's pin: the file stays \`in_use\` forever (until process restart), permanently disabling LRU eviction for it.

## Affected paths

- **\`/leave\`** (\`commands/leave.py\`) — calls \`player.queue.clear()\` after \`player.stop()\`. The currently-playing track gets a STOPPED event so its pin is released; **queued-but-never-played tracks leak**.
- **\`EventHandler.on_websocket_closed\` (4014, voice channel disconnected)** — same shape: bot removed from voice, queue cleared, queued tracks leak. Pre-existing.
- **\`EventHandler.on_node_disconnected\`** — same shape. Pre-existing.

Originally surfaced in the \`/security-review\` of #23 (M-1):

&gt; The \`/leave\` (\`commands/leave.py:99-103\`) calls \`player.queue.clear()\` without releasing \`audio_cache\` pins for queued-but-never-played tracks. Each leaked pin permanently disables LRU eviction for that file. […] DoS via slow disk-fill; recoverable on bot restart.

## Severity

Medium-impact bug. Recoverable on bot restart. Triggerable by users via \`/leave\`. Practical exploitability is bounded by the cache size cap, but a noisy guild can convert most of \`RYZIC_CACHE_DIR\` into permanently-pinned files in a single session.

## Suggested fix

Centralise queue-clear to release pins for every \`AudioTrack\` in the queue before clearing. Likely a single helper on the audio glue (\`lavalink_glue.py\`) — \`async def clear_queue_releasing(player)\` — used by all three call sites.

Add a unit test against \`tests/_command_helpers.FakePlayer\` that records cache releases on queued (not just playing) tracks.

## Out of scope for #23

#23 introduced the first **user-triggered** path with this bug, but the underlying pattern is older. Fixing only \`/leave\` would leave the websocket / node-disconnect paths broken. Better to fix the pattern in one PR after #23 lands.

riohno:
## Summary

Fixes #24 — `audio_cache` pin leak on the three queue-clear paths that drop unplayed tracks without firing `TrackEndEvent`.

`EventHandler._release_track` releases pins on `TrackEndEvent` (FINISHED / REPLACED / STOPPED / LOAD_FAILED), but tracks that sat in `player.queue` and were dropped without ever playing never see that event. Their pins stay in `audio_cache._in_use` forever — `LRU` eviction permanently skips them. A noisy guild can convert most of `RYZIC_CACHE_DIR` into permanently-pinned files in a single session, recoverable only on bot restart.

## Design

Centralise queue-clear behind `clear_queue_releasing(player)` in `lavalink_glue.py`. The helper iterates `player.queue` calling the existing `_release_track` (so the path-stem-from-identifier extraction, `None`-track guard, missing-cache guard, and exception swallowing are all reused), then empties the queue.

Three call sites switch from `player.queue.clear()` to `await clear_queue_releasing(player)`:

1. `commands/leave.py` — `/leave` (user-triggered, the path that surfaced this in #23's security review).
2. `lavalink_glue.EventHandler.on_websocket_closed` (Discord 4014: voice channel kicked / channel deleted).
3. `lavalink_glue.EventHandler.on_node_disconnected` (lavalink node drop).

Out of scope per the issue: TTL on the cache; touching `_release_track`; touching `/skip` (which fires REPLACED/STOPPED correctly).

## Files changed

- `src/ryzic/lavalink_glue.py` — add `clear_queue_releasing(player)`; switch the two `EventHandler` queue-clear sites to it.
- `src/ryzic/commands/leave.py` — switch the `/leave` queue-clear to the helper.
- `tests/test_leave_command.py` — add a `_RecordingCache`, extend the autouse fixture to install/clear it, and add `test_leave_releases_audio_cache_pins_for_queued_tracks` asserting both queued tracks' video_ids get released on `/leave`.
- `tests/test_lavalink_bridge.py` — add `_QueuePlayer` stand-in plus tests for the helper itself (happy path, empty queue, missing-identifier skip), the websocket-closed 4014 path (releases) vs non-4014 (no-op), and the node-disconnected path across multiple players.

## Test plan

- [x] `uv run pytest -q --cov-fail-under=80` — 312 passed, coverage 90.71%

Closes #24.

### DIFF
--- a/src/ryzic/commands/leave.py
+++ b/src/ryzic/commands/leave.py
@@ -52,9 +52,10 @@ async def _handle_leave(ctx: lightbulb.Context) -> None:
 
     # Stop first so Lavalink halts the stream cleanly; clearing the queue
     # before update_voice_state guarantees a follow-up auto-advance can
-    # not race a disconnect.
+    # not race a disconnect. Use the releasing helper so audio_cache pins
+    # for queued-but-never-played tracks don't leak (issue #24).
     await player.stop()
-    player.queue.clear()
+    await lavalink_glue.clear_queue_releasing(player)
 
     bot = cast(hikari.GatewayBot, ctx.client.app)
     await bot.update_voice_state(guild_id, None)
--- a/src/ryzic/lavalink_glue.py
+++ b/src/ryzic/lavalink_glue.py
@@ -223,10 +223,18 @@ async def clear_queue_releasing(player: lavalink.DefaultPlayer) -> None:
     leave their pins held forever. Centralising the release walk here
     keeps the three callers (``/leave``, voice 4014 disconnect, node
     disconnect) honest about the contract.
+
+    Snapshot the queue and clear it BEFORE awaiting any release: a
+    concurrent ``/play`` (or other ``player.queue.add`` caller) that
+    lands between ``await`` boundaries on the snapshot would otherwise
+    have its newly-queued track silently wiped by the trailing
+    ``queue.clear()``, leaking its pin. Clear-then-release leaves a
+    fresh queue for any racing producer.
     """
-    for track in player.queue:
-        await _release_track(track)
+    tracks = list(player.queue)
     player.queue.clear()
+    for track in tracks:
+        await _release_track(track)
 
 
 def _safe_error_text(text: str | None) -> str:
--- a/src/ryzic/lavalink_glue.py
+++ b/src/ryzic/lavalink_glue.py
@@ -215,6 +215,20 @@ async def _release_track(track: lavalink.AudioTrack | None) -> None:
         _log.exception("failed to release audio cache entry %s", video_id)
 
 
+async def clear_queue_releasing(player: lavalink.DefaultPlayer) -> None:
+    """Release audio cache pins for every queued track, then clear the queue.
+
+    ``TrackEndEvent`` fires (and ``_release_track`` runs) only for tracks
+    that actually played; queue-clear paths that drop unplayed tracks
+    leave their pins held forever. Centralising the release walk here
+    keeps the three callers (``/leave``, voice 4014 disconnect, node
+    disconnect) honest about the contract.
+    """
+    for track in player.queue:
+        await _release_track(track)
+    player.queue.clear()
+
+
 def _safe_error_text(text: str | None) -> str:
     """Sanitise server-supplied error text before posting to a public channel.
 
@@ -327,7 +341,7 @@ class EventHandler:
         )
         if event.code != 4014:
             return
-        cast(lavalink.DefaultPlayer, event.player).queue.clear()
+        await clear_queue_releasing(cast(lavalink.DefaultPlayer, event.player))
         _cancel_auto_leave(guild_id)
         _reset_voice_ready(guild_id)
         await _send_to_last_play_channel(
@@ -348,7 +362,7 @@ class EventHandler:
         notified: set[int] = set()
         for player in list(client.player_manager.values()):
             guild_id = player.guild_id
-            cast(lavalink.DefaultPlayer, player).queue.clear()
+            await clear_queue_releasing(cast(lavalink.DefaultPlayer, player))
             _cancel_auto_leave(guild_id)
             _reset_voice_ready(guild_id)
             if guild_id in notified:
--- a/tests/_command_helpers.py
+++ b/tests/_command_helpers.py
@@ -104,6 +104,22 @@ class FakeAudioTrack:
     extra: dict[str, Any] = field(default_factory=dict)
 
 
+class RecordingCache:
+    """Captures ``release`` calls so tests can assert pin release on queue clear.
+
+    Used by tests that exercise ``lavalink_glue.clear_queue_releasing`` and
+    its callers (``/leave``, voice 4014 disconnect, node disconnect) — the
+    real :class:`~ryzic.audio_cache.AudioCache` is sqlite-backed and
+    overkill for assertions about the release walk's contract.
+    """
+
+    def __init__(self) -> None:
+        self.released: list[str] = []
+
+    async def release(self, video_id: str) -> None:
+        self.released.append(video_id)
+
+
 class FakePlayer:
     def __init__(self, guild_id: int) -> None:
         self.guild_id = guild_id
--- a/tests/test_lavalink_bridge.py
+++ b/tests/test_lavalink_bridge.py
@@ -19,6 +19,7 @@ import lavalink
 import pytest
 
 from ryzic import lavalink_glue
+from tests._command_helpers import RecordingCache
 
 
 @dataclass
@@ -575,16 +576,6 @@ def test_is_valid_discord_endpoint_allowlist() -> None:
 # ---------------------------------------------------------------------------
 
 
-class _RecordingCache:
-    """Captures release() calls so tests can assert the integration."""
-
-    def __init__(self) -> None:
-        self.released: list[str] = []
-
-    async def release(self, video_id: str) -> None:
-        self.released.append(video_id)
-
-
 @dataclass
 class _IdTrack:
     title: str = "song"
@@ -621,7 +612,7 @@ async def test_track_end_releases_audio_cache_pin() -> None:
     """
     from ryzic import audio_cache
 
-    fake = _RecordingCache()
+    fake = RecordingCache()
     audio_cache.set_audio_cache(cast(audio_cache.AudioCache, fake))
     try:
         bot = _FakeApp()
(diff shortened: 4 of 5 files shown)


riohno (commented):
## /simplify — PR #28 (audio_cache pin leak fix)

**Verdict:** small wins

**Estimated trim:** ~15 lines

### Wins (in order of value)

1. **De-dup `_RecordingCache`** — `tests/test_lavalink_bridge.py:578` and `tests/test_leave_command.py:22` now hold an identical 6-line class (same fields, same `release` signature, same captured list). With two consumers it crosses the bar suggested in the project's own helper file (`tests/_command_helpers.py`'s docstring already says "centralising them here keeps the … test files focused"). Move the canonical version to `tests/_command_helpers.py` and import it from both places. Saves ~10 lines and keeps a third future caller honest. (~10 LOC)

2. **`test_clear_queue_releasing_skips_tracks_without_identifier` is largely redundant** — `_release_track`'s missing-identifier guard is already covered by `test_release_is_noop_for_none_track` and the `_RecordingCache`-driven happy-path tests; asserting the same guard through the wrapper just re-tests the underlying function. The empty-queue test is a real boundary (helper-specific, no underlying analogue), so keep it. Drop only the missing-identifier test. (~18 LOC saved, but it does pin the wrapper's inheritance of the contract — borderline; defensible to keep.)

3. **Helper docstring is one paragraph long for a 3-line function** — `lavalink_glue.py:218-226`. The \"why\" (TrackEndEvent doesn't fire for unplayed tracks) is the load-bearing bit; the parenthetical \"(/leave, voice 4014, node disconnect)\" is a maintenance hazard if a fourth caller appears. Trim to two sentences: contract + the \"only fires for played tracks\" rationale. (~3 LOC)

### Items considered, no win

- **Helper sync vs async:** `_release_track` awaits `cache.release(...)`, so the helper must be async. No change.
- **Inlining the helper:** 3 call sites + identical logic = DRY win, helper stays.
- **Empty-queue test:** valuable defensive boundary, keep.
- **Comment in `commands/leave.py`:** the \"audio_cache pins ... don't leak (issue #24)\" line is a real WHY referencing the issue — keep.
- **Comment block in `test_leave_releases_audio_cache_pins_for_queued_tracks` (L140-142)** explains why the currently-playing track is excluded from the assertion — non-obvious test boundary, worth keeping.
- **`_QueuePlayer` fixture (test_lavalink_bridge.py):** scope-local to that file's queue-clear tests; doesn't meet the >1-file bar for promotion to `_command_helpers.py`.
- **Per-test `from ryzic import audio_cache`:** matches the file's existing pattern (the pre-existing `_release_track` tests do the same); leaving alone for consistency.

riohno (commented):
## /security-review — PR #28 (audio_cache pin leak fix)

**Verdict:** ship as-is

**HIGH-severity findings:** 0

### Findings

**[LOW] queue mutation during sweep is unsynchronized** — `lavalink_glue.py:226-229`.
Threat model: a `/play` racing a `/leave` can `player.add()` a freshly-loaded track between the helper's `await _release_track(...)` boundaries, so the new track gets pin-released before it ever plays — exposing its file to LRU eviction. Counter underflow is not possible (the `release()` guard handles it), and a follow-up `TrackEndEvent` is a safe no-op. The pre-existing race window is the same one the original `queue.clear()` had; the PR does not introduce it, only changes its observable consequence from "leak" to "early release". Fix (optional, defense-in-depth): snapshot via `tracks = list(player.queue); player.queue.clear()` before the await loop so iteration runs against a private list. Worth ~2 lines and removes the only window where eviction could touch a soon-to-play file.

### Items checked, no finding

- **Counter underflow on double-release.** `audio_cache.release` (line 231) explicitly guards `if self._in_use[video_id] <= 0` and `Counter[key]` returns 0 without auto-inserting, so a double-release for the same `video_id` (helper sweep + a racing `TrackEndEvent`) is a clean no-op. No path to a negative counter.
- **Per-track exceptions in the sweep.** `_release_track` already wraps `await cache.release(...)` in try/except (line 210-215), so a misbehaving cache cannot abort the helper mid-iteration. The helper correctly relies on that contract instead of redundantly re-wrapping.
- **Test fake matches production contract.** `_RecordingCache.release(video_id: str)` (in both `test_lavalink_bridge.py` and `test_leave_command.py`) is signature-compatible with `AudioCache.release`. Production's `release` does no validation either (pure Counter bookkeeping), so the fake doesn't conceal any rejected-input behaviour.
- **Untrusted-input boundary on `identifier`.** `track.identifier` is populated by `LocalAudioSourceManager` from the file path `_load_one` passes to `node.get_tracks(str(path))`, which is built from `cache_root + video_id` after `validate_video_id` (`^[A-Za-z0-9_-]{6,20}$`). `Path(identifier).stem` reduces to that validated stem. `cache.release` performs no filesystem ops on the value, so even a hypothetical injected stem could only no-op against the in-memory Counter.
- **Empty / missing identifier.** `if not identifier: return` in `_release_track` (line 202) means `Path("").stem == ""` never reaches `cache.release`. The dedicated test (`test_clear_queue_releasing_skips_tracks_without_identifier`) covers this.
- **Helper used uniformly across all three call sites.** `/leave`, voice 4014, and node disconnect are the only `queue.clear()` paths in the repo (`grep -rn queue.clear src/`); each was switched. `/skip` is intentionally untouched per the PR description (REPLACED/STOPPED fire `TrackEndEvent`).

riohno (commented):
## /review — PR #28 (audio_cache pin leak fix)

**Verdict:** ship as-is

### Findings

None blocking. Two minor observations recorded as "items checked" below rather than findings, since neither is a regression vs. the pre-PR state.

### Items checked, no finding

- **Helper correctness:** `clear_queue_releasing` iterates `player.queue` (a plain `List[AudioTrack]` from `lavalink/player.py:122`), releases per-track, then `.clear()`s. Order is correct (release THEN clear).
- **Empty queue / missing identifier:** `_release_track` short-circuits on `track is None`, missing cache singleton, and falsy `identifier`; the helper's loop is a no-op for empty queues. Covered by `test_clear_queue_releasing_handles_empty_queue` and `test_clear_queue_releasing_skips_tracks_without_identifier`.
- **Per-track exception swallowing:** `_release_track` already swallows `cache.release` exceptions (`lavalink_glue.py:212-215`). With the production cache, `AudioCache.release` is `async` but contains no `await`s, so the loop does not yield to other tasks during iteration — no concurrent-mutation window in practice.
- **Reuse of `_release_track`:** signature `(track: AudioTrack | None) -> None`, side effect = decrement `audio_cache._in_use[video_id]`. The path-stem extraction (`Path(identifier).stem`) is the same key the queued track was pinned with in `get_or_download`, so the release lands on the right counter. Confirmed against `test_release_with_path_identifier_decrements_real_cache_pin`.
- **All 3 call sites switched:** `commands/leave.py:57`, `lavalink_glue.py:344` (websocket 4014), `lavalink_glue.py:365` (node disconnect). `grep -rn "queue.clear" src/` returns zero hits after the patch.
- **4014 gating preserved:** `on_websocket_closed` still early-returns on `event.code != 4014` before calling the helper (`lavalink_glue.py:342-344`); behaviour is unchanged for transient codes (1006 etc.). Covered by `test_websocket_closed_non_4014_does_not_touch_queue`.
- **Multi-player node-disconnect:** `on_node_disconnected` snapshots `player_manager.values()` via `list(...)` before iterating, so concurrent player_manager mutation during `await` cannot corrupt the loop. Covered by `test_node_disconnected_releases_queued_pins_for_every_player` (two guilds, three tracks).
- **Test assertion strength:** all six new tests assert against `_RecordingCache.released` (the actual cache release calls), not just absence of exceptions.
- **Race on `/leave` between `player.stop()` and queue clear:** pre-existing; unchanged by this PR. `EndReason.STOPPED.may_start_next()` is `False` (`lavalink/server.py:229`), so `stop()` does not auto-advance the queue. A concurrent `/play` task that reaches `player.add()` between the two awaits is theoretically possible but pre-existing and out of scope.
- **Out-of-scope paths intentionally not touched:** `_on_guild_leave` calls `player_manager.destroy(guild_id)` which doesn't release pins either; per the PR scoping (issue #24's three paths only) this is correctly left alone.
- **CI:** `uv run pytest -q` → 312 passed, coverage 91%; `ruff check`, `ruff format --check`, `ty check` all clean.

PR #37, opened 2026-05-03
feat(deploy): pull-only compose.yaml for v0.1.0 GHCR consumers

riohno:
## Summary

Restructures the deploy story so the canonical "self-hostable" UX is fetch-three-files-and-up — operators don't need a source clone to run the published GHCR image. With v0.1.0 about to ship the first GHCR build (release-please PR #30), the README-default flow should match what users actually want: pull a published image, not maintain a working tree.

## What changed

- **`compose.yaml`** swaps `build: .` for `image: ghcr.io/verylongorgnamesuchwow/ryzic:0.1`. Pinned to the **minor tag** (not `:latest`) for predictable updates: `docker compose pull` picks up 0.1.x patches automatically; bumping to a future `:0.2` is an explicit manual edit, since pre-1.0 minor bumps may include breaking changes ([SEMVER.md](SEMVER.md)).
- **`compose.dev.yaml` (new)** is a thin overlay re-adding `build: .` for contributors iterating against a working tree:
  ```
  docker compose -f compose.yaml -f compose.dev.yaml up --build
  ```
  Sets `image: ryzic:dev` to avoid cache collisions with the pulled GHCR image.
- **README** install steps now lead with `mkdir ryzic && curl compose.yaml + .env.example + lavalink/application.yml` (the three files an operator actually needs on disk). The `git clone` flow moves under **Development**, alongside the dev-overlay command.
- **`release-dry-run.yml`** also watches `compose.dev.yaml` so changes to it still gate the release path.
- **`.github/labeler.yml`** tags `compose.dev.yaml` under `area:deploy`.

## Design choice (Option A vs B)

The brief offered two shapes: (A) make `compose.yaml` the pull-only one and add a dev overlay, or (B) keep `compose.yaml` as dev-shaped and add a separate `compose.prod.yaml`. **This PR takes A.**

Rationale:
- The pull-only file is what "just want to run it" operators want to be the default they read on the README.
- The dev overlay only needs to add `build:` (and disambiguate the local image tag) — minimal drift surface, since the lavalink/healthcheck/volume/cache stanzas live exclusively in `compose.yaml` and are inherited.
- Option B would require the prod file to either duplicate those stanzas (drift risk) or use the same overlay mechanism in reverse (operators would need to remember to invoke two files). Option A keeps the operator path single-file.

## Image tag pinning

`release.yml`'s `docker/metadata-action` step already emits `:X.Y` from the resolved tag, so `:0.1` automatically points at the latest 0.1.x patch on every release-please cut — no `release-please-config.json` change needed. When 0.2 ships, `:0.1` will freeze (release-please doesn't repoint old minor tags); operators stay on 0.1.x until they manually edit `compose.yaml` to `:0.2`. That's the intended pre-1.0 upgrade ceremony.

## Coordination notes

- `ryzic-cleanup`'s in-flight upgrade-snippet PR adds a `docker compose pull && up -d` section. The pull-only compose here is the foundation that snippet builds on; either PR can land first. If theirs lands first I'll rebase and drop the (smaller) upgrade hint already in §3 to avoid duplication; if mine lands first their snippet slots in cleanly.
- README is also touched by `ryzic-cleanup`'s items 3/4/5. Likely a textual conflict on the install section; whichever lands second rebases.

## Verification

- `uv run ruff check` — passes
- `uv run ruff format --check` — passes
- `uv run ty check` — passes
- `uv run pytest -q --cov-fail-under=80` — 314 passed, coverage 90.91%
- `python3 -c "import yaml; yaml.safe_load(open('<file>'))"` for compose.yaml, compose.dev.yaml, release.yml, release-dry-run.yml, labeler.yml — all OK
- `podman-compose -f compose.yaml config` — resolves, image is `ghcr.io/verylongorgnamesuchwow/ryzic:0.1`
- `podman-compose -f compose.yaml -f compose.dev.yaml config` — resolves, overlay correctly adds `build:` and overrides `image: ryzic:dev`

## Test plan

- [ ] CI green (lint/type/test/coverage + release-dry-run on the new compose.dev.yaml)
- [ ] After v0.1.0 ships, smoke-test the pull-only README flow end-to-end against the live image

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -37,6 +37,7 @@
           - "Dockerfile"
           - ".dockerignore"
           - "compose.yaml"
+          - "compose.dev.yaml"
           - "lavalink/**"
           - "tests/integration/**"
 
--- a/.github/workflows/release-dry-run.yml
+++ b/.github/workflows/release-dry-run.yml
@@ -14,6 +14,7 @@ on:
       - Dockerfile
       - .dockerignore
       - compose.yaml
+      - compose.dev.yaml
       - lavalink/application.yml
       - .github/workflows/release.yml
       - .github/workflows/release-dry-run.yml
--- a/README.md
+++ b/README.md
@@ -36,12 +36,17 @@ A self-hostable Discord music bot. Plays YouTube audio in voice channels via [La
    - **Do not pick Administrator.** A music bot has no business with admin rights — it's a security smell and many server owners refuse such invites.
 5. Open the generated URL and invite the bot to your server.
 
-### 2. Clone and configure
+### 2. Fetch the deploy files
+
+You only need three files on disk — the published [`ghcr.io/verylongorgnamesuchwow/ryzic`](https://github.com/VeryLongOrgNameSuchWow/ryzic/pkgs/container/ryzic) image carries the bot itself, so there's no source tree to maintain.
 
 ```bash
-git clone https://github.com/VeryLongOrgNameSuchWow/ryzic.git
-cd ryzic
-cp .env.example .env
+mkdir ryzic && cd ryzic
+curl -fsSLO https://raw.githubusercontent.com/VeryLongOrgNameSuchWow/ryzic/main/compose.yaml
+curl -fsSLO https://raw.githubusercontent.com/VeryLongOrgNameSuchWow/ryzic/main/.env.example
+mkdir lavalink && curl -fsSL -o lavalink/application.yml \
+  https://raw.githubusercontent.com/VeryLongOrgNameSuchWow/ryzic/main/lavalink/application.yml
+mv .env.example .env
 ```
 
 Edit `.env` and paste the token into `DISCORD_BOT_TOKEN=`.
@@ -58,11 +63,13 @@ RYZIC_GUILD_IDS=123456789012345678
 docker compose up -d
 ```
 
-The first boot pulls the Lavalink image and downloads the `youtube-source` plugin (~30s). Tail the logs to confirm both services are healthy:
+The first boot pulls the ryzic + Lavalink images and downloads the `youtube-source` plugin (~30s). Tail the logs to confirm both services are healthy:
 
 ```bash
 docker compose logs -f
 ```
+
+`compose.yaml` pins the bot to the `:0.1` minor tag — `docker compose pull && docker compose up -d` picks up patch releases on the 0.1.x line. Bumping to a future `:0.2` is a manual edit, since pre-1.0 minor bumps may include breaking changes (see [SEMVER.md](SEMVER.md)).
 
 ### 4. Try it
 
@@ -109,10 +116,20 @@ All configuration is via environment variables (read from `.env` by `docker comp
 You don't need Docker to hack on the code — only to run a real Lavalink (which the integration tests spin up themselves via testcontainers).
 
 ```bash
+git clone https://github.com/VeryLongOrgNameSuchWow/ryzic.git
+cd ryzic
 uv sync
 uv run pytest -q
 uv run python -m ryzic    # requires DISCORD_BOT_TOKEN + a reachable Lavalink
 ```
+
+To run the local working tree under compose instead of the published GHCR image, layer the dev overlay on top of `compose.yaml`:
+
+```bash
+docker compose -f compose.yaml -f compose.dev.yaml up --build
+```
+
+The overlay swaps `image: ghcr.io/...` for `build: .` so iteration doesn't require pushing to GHCR.
 
 Run the same lint/type/test suite CI runs before opening a PR — see [CONTRIBUTING.md § Pull requests](CONTRIBUTING.md#pull-requests) for the canonical command. [docs/manual-smoke-tests.md](docs/manual-smoke-tests.md) is the end-to-end checklist run before each release.
 
--- a/README.md
+++ b/README.md
@@ -73,8 +73,6 @@ The first boot pulls the ryzic + Lavalink images and downloads the `youtube-sour
 docker compose logs -f
 ```
 
-`compose.yaml` pins the bot to the `:0.1` minor tag — `docker compose pull && docker compose up -d` picks up patch releases on the 0.1.x line. Bumping to a future `:0.2` is a manual edit, since pre-1.0 minor bumps may include breaking changes (see [SEMVER.md](SEMVER.md)).
-
 ### 4. Try it
 
 Join a voice channel, then in any text channel the bot can see:
@@ -115,7 +113,9 @@ docker compose pull
 docker compose up -d
 ```
 
-The image is pinned to `:latest` by default; pin to a specific `:vX.Y.Z` in your `compose.yaml` if you want manual control over upgrades. The audio + playlist cache (the `RYZIC_CACHE_DIR` bind mount) and the SQLite database inside it persist across upgrades — `docker compose pull` only replaces the image, not the volume.
+`compose.yaml` pins the bot to the `:0.1` minor tag, so `docker compose pull` picks up 0.1.x patch releases automatically. Bumping to a future `:0.2` is a manual edit — pre-1.0 minor bumps may include breaking changes (see [SEMVER.md](SEMVER.md)). Pin to a specific `:vX.Y.Z` if you want full manual control.
+
+The audio + playlist cache (the `RYZIC_CACHE_DIR` bind mount) and the SQLite database inside it persist across upgrades — `docker compose pull` only replaces the image, not the volume.
 
 ## Self-hoster considerations
 
--- a/compose.dev.yaml
+++ b/compose.dev.yaml
@@ -1,0 +1,16 @@
+# Compose overlay for local development against a working tree.
+#
+# Usage:
+#   docker compose -f compose.yaml -f compose.dev.yaml up --build
+#
+# This overlay replaces the published GHCR `image:` from compose.yaml with a
+# `build:` from the local Dockerfile, so contributors can iterate without
+# pushing to GHCR. Operators just running the bot don't need this file.
+
+services:
+  ryzic:
+    build: .
+    # The image: from compose.yaml is still inherited; setting it to an
+    # explicit dev tag prevents the local build from cache-colliding with a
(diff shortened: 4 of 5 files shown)


riohno (commented):
## /simplify — PR #37 (pull-only compose.yaml)

**Verdict:** ⚠️ small wins

**Estimated trim:** ~5 lines

### Wins (in order of value)

1. **dev-overlay header is now redundant with README** — `compose.dev.yaml:1-8`. The 8-line preamble (purpose + `docker compose -f ... up --build` invocation + rationale) restates what `README.md:126-132` now says in prose, in the place an operator/contributor would actually look first. The README is load-bearing for this; the YAML comment becomes a second source of truth that can drift. Suggest collapsing to a one-liner pointer:
    ```yaml
    # Dev overlay for local working-tree builds. See README "Development".
    ```
    The `image: ryzic:dev` cache-collision comment (lines 13-15) is non-obvious WHY and worth keeping. (~5 lines saved)

### Items considered, no win

- **Compose duplication between files** — already minimal. `compose.dev.yaml` only adds `build:` + a dev tag; lavalink/healthcheck/volume/cache stanzas live exclusively in `compose.yaml` and are inherited. Zero duplication, no `extends:`/anchors needed.
- **Option A vs B choice** — A (canonical + 2-line dev overlay) is the right call given the inheritance behaviour above; B would have forced duplicating the lavalink stanza into a prod file or required operators to remember a two-file invocation.
- **README dev-flow vs pull-only-flow cohesion** — distinct sections (§2 = `mkdir + curl`, Development = `git clone` + overlay command). No overlap.
- **`compose.yaml:3-5` minor-tag comment** — non-obvious WHY (manual edit ceremony for `:0.2`), survives the "would a reader infer this from the YAML?" test. Keep.
- **README:72 minor-tag paragraph vs `compose.yaml` comment** — intentional layering (operators read README, not YAML); not worth de-duplicating.
- **labeler.yml / release-dry-run.yml additions** — 1 line each, minimum viable.
- **No speculative "support both X and Y" patterns.**

riohno (commented):
## /security-review — PR #37 (pull-only compose.yaml)

**Verdict:** ✅ ship as-is (one MEDIUM nit pre-existing, called out for awareness)
**HIGH-severity findings:** 0

### Findings

**[MEDIUM] Lavalink `LAVALINK_PASSWORD` defaults to `youshallnotpass`** — `compose.yaml:24` (pre-existing, not introduced by this PR). An operator who never reads `.env.example` and just runs `docker compose up -d` would launch Lavalink with the well-known default password. Threat model: limited because no `ports:` block exposes 2333 (Lavalink is reachable only on the compose-internal bridge network), so an attacker would need an existing foothold on the host to abuse it. Worth a follow-up: drop the `:-youshallnotpass` default so compose refuses to start without a password set, mirroring how `application.yml:15` already requires `${LAVALINK_SERVER_PASSWORD}` with no literal fallback. Out of scope for this PR — flag for a separate change.

**[INFO] Composes well with PR #38 (build-provenance attestation)** — the README shipped here teaches operators the `curl compose.yaml | docker compose pull` flow; PR #38 adds SLSA build provenance + sigstore attestation pushed to GHCR alongside the image. After both land, an operator who wants to verify what they pulled has a path: `gh attestation verify oci://ghcr.io/verylongorgnamesuchwow/ryzic@<digest> --repo VeryLongOrgNameSuchWow/ryzic`. Consider a follow-up README note pointing at `gh attestation verify` once #38 merges, so the trust chain is discoverable. Not a blocker for #37.

### Items checked, no finding

- **ryzic image tag pinning** — `:0.1` (minor) is the right pre-1.0 compromise: `docker compose pull` picks up 0.1.x patches without surprise major/minor jumps. `release.yml`'s `docker/metadata-action` correctly emits the `{major}.{minor}` tag from the resolved release tag, and `:latest` is gated to `push` events only (workflow_dispatch rebuilds of old tags won't move `:latest` backward).
- **Lavalink image pinning** — `ghcr.io/lavalink-devs/lavalink:4.2.2` is a full minor.patch pin (stricter than ryzic's own minor pin). `docker compose pull` won't auto-roll Lavalink; operator-controlled bump.
- **Privileged operations** — no `cap_add`, `privileged: true`, `security_opt`, `sysctls`, `cgroup_parent`, or `userns_mode` in either compose file. Container runs as UID 1001 unprivileged user per `Dockerfile:33-40`.
- **Network exposure** — no `ports:` block in either compose. ryzic and Lavalink only talk over the compose-internal bridge; nothing inbound from the host.
- **Required env vars** — `DISCORD_BOT_TOKEN` has no insecure default: `src/ryzic/config.py:80` uses `_require()` and raises `ConfigError` if unset. `.env.example:5` ships it as empty (`DISCORD_BOT_TOKEN=`), which fails fast rather than silently using `""`.
- **Compose `secrets:`** — not used. Token flows via `env_file: .env`, which is `.gitignore`d. No bind-mount of operator credentials.
- **Mount paths** — `cache:` and `lavalink_plugins:` are named volumes (Docker-managed, not bind mounts to user-writable host paths). The single bind `./lavalink/application.yml:/opt/Lavalink/application.yml:ro` is read-only and points at the operator's own config that the README tells them to fetch — appropriate for Lavalink's config-file-only loading model.
- **`.env.example` shape** — still shipped, and the new README flow (`mkdir ryzic && curl ... && mv .env.example .env`) correctly fetches it as one of the three required files. Operators following the pull-only path get the full env-var contract.
- **`compose.dev.yaml` overlay** — `image: ryzic:dev` correctly disambiguates the local build from a previously-pulled GHCR image of the same name; no risk of cache collision pushing a stale local build to a pulled tag.

riohno (commented):
## /review — PR #37 (pull-only compose.yaml)

**Verdict:** ⚠️ minor revisions (one blocking README contradiction; rest are nits)

### Findings

**[BLOCKING] README: ':0.1' pin contradicts the merged ':latest' Upgrading section** — `README.md:72` (this PR) vs `README.md:118` (from #35, already on `main`).

The PR branched off `9e8584f` — before #35 (`9582724`) merged the new `## Upgrading` section. That section claims *"The image is pinned to `:latest` by default; pin to a specific `:vX.Y.Z` in your `compose.yaml` if you want manual control over upgrades."* — which was already wrong relative to a not-yet-existing pin, and becomes actively contradictory once this PR lands and pins to `:0.1`. The merge auto-resolves cleanly (`mergeStateStatus: BEHIND` but `mergeable: MERGEABLE`), so no conflict marker will catch it.

After fast-forward, `README.md` will tell users two different things about the default tag in two adjacent sections. Concretely:

```
L72:  `compose.yaml` pins the bot to the `:0.1` minor tag — `docker compose pull && docker compose up -d` …
L118: The image is pinned to `:latest` by default; pin to a specific `:vX.Y.Z` in your `compose.yaml` if you want manual control over upgrades.
```

**Fix:** rebase on `main` and rewrite the Upgrading section to match the new reality — drop the `:latest` claim, point at the `:0.1` pin from §3, and reuse the SEMVER.md cross-link rather than duplicating the "pin to `:vX.Y.Z`" guidance. While there, the §3 paragraph (L72) and the Upgrading section largely overlap; one of them can shrink to a one-liner referencing the other (the brief explicitly asked for a cross-link to the upgrade snippet from §3, which is also currently missing).

**[NIT] `.dockerignore` doesn't exclude `compose.dev.yaml`** — `.dockerignore:30-32`. `compose.yaml`, `docker-compose.yaml`, `docker-compose.yml` are all excluded with the comment *"Compose stack and lavalink config aren't part of the bot image"*; the new dev overlay belongs in that list for symmetry. Won't materially bloat the image (16 lines), but the intent of that block is "no compose files in the runtime image". **Fix:** add `compose.dev.yaml` to the excluded list.

**[NIT] Pull-only README flow has no upgrade cross-link from §3** — `README.md:72`. The brief called out cross-linking §3 to PR #35's `## Upgrading` section so the README stays a single coherent narrative. The §3 paragraph re-states the upgrade command inline rather than linking, which both duplicates content and obscures the fact that the cache + DB persist across upgrades (the key reassurance #35 added). **Fix:** in §3, replace the inline `docker compose pull && docker compose up -d` example with a one-line pointer to `## Upgrading`, and let §Upgrading own the full story (after it's been rewritten per the blocking finding).

### Items checked, no finding
- Image tag pinned to minor (`:0.1`), not `:latest` — matches brief.
- `release.yml`'s `docker/metadata-action` already emits `{{major}}.{{minor}}` from the resolved tag, so `:0.1` is published on every 0.1.x release without config change.
- `release-please-config.json` has no `extra-files` entry — `:0.1` in `compose.yaml` will NOT be auto-bumped, which is the intended pre-1.0 ceremony per `SEMVER.md` (operator action required to move to `:0.2`). Behaviour matches the PR description.
- All `compose.yaml` stanzas preserved verbatim — depends_on, env_file, environment (`LAVALINK_HOST`/`LAVALINK_PORT`/`RYZIC_CACHE_DIR`), volumes, restart, lavalink service block (image pin, healthcheck, plugins/cache mounts, password env), top-level `volumes: { cache, lavalink_plugins }`. Only delta is `build: .` → `image: ghcr.io/...:0.1` plus the comment.
- Cache volume name (`cache`) unchanged — in-place upgrade for existing self-hosters, no fresh wipe.
- `compose.dev.yaml` overlay sets `image: ryzic:dev` to avoid cache collision with the pulled GHCR image — sound reasoning, prevents a subtle "why is my code not loading" footgun.
- Backward compat: dev path still works via `compose.yaml + compose.dev.yaml` overlay; the existing `git clone` flow is preserved under §Development.
- `compose.yaml` and `compose.dev.yaml` both parse as valid YAML; PR author already ran `podman-compose ... config` on both combinations; CI's `docker-build` + `dry-run` jobs pass.
- `.github/labeler.yml` adds `compose.dev.yaml` under `area:deploy` — consistent with `compose.yaml` placement.
- `.github/workflows/release-dry-run.yml` paths-filter includes `compose.dev.yaml` — release path stays gated on overlay changes.
- `lavalink/application.yml` exists at the README-curl'd path; bind mount target unchanged.
- README design choice (Option A: pull-only as default, dev as overlay) is consistent across `compose.*.yaml` + README + labeler + release-dry-run — no drift between files.

PR #38, opened 2026-05-03
feat(release): SLSA build provenance attestation on GHCR push

riohno:
## Summary

Signs every published GHCR image with SLSA build provenance via [`actions/attest-build-provenance`](https://github.com/actions/attest-build-provenance), giving consumers cryptographic linkage from `:tag` → digest → source commit → builder identity. Brings ryzic to parity with kbmcp, which gets the equivalent free via `pypa/gh-action-pypi-publish`'s default `attestations: true`.

## What changed

- New `permissions:` on the `release` workflow:
  - `id-token: write` — mints the OIDC token sigstore uses to identify the builder.
  - `attestations: write` — writes the generated SLSA provenance to the repo's attestation store.
- Added `id: build` to the existing `docker/build-push-action` step so its `outputs.digest` is referenceable.
- Added `actions/attest-build-provenance@a2bbfa25375fe432b6a289bc6b6cd05ecd0c4c32` (v4.1.0, SHA-pinned) after the build/push step, with:
  - `subject-name: ghcr.io/${{ github.repository }}`
  - `subject-digest: ${{ steps.build.outputs.digest }}`
  - `push-to-registry: true` — uploads the attestation as an OCI artifact alongside the image, so anyone with `gh attestation verify oci://ghcr.io/<repo>@<digest> --repo <repo>` can verify without API access.

## Verifying a release

After the next release-please tag cuts, consumers can run:

```bash
gh attestation verify oci://ghcr.io/verylongorgnamesuchwow/ryzic@<digest> --repo VeryLongOrgNameSuchWow/ryzic
```

…to confirm the image was built from this repository, by this workflow, on this commit. The existing softprops-driven release-notes section composes nicely: users get the digest from the GitHub Release body, then verify it against the attestation in one command.

## CI behavior

**This step will not produce attestations during PR CI** — `release.yml` only runs on tag push or `workflow_dispatch`. PRs don't fire it, so there's no PR-time check to gate on. The first real validation will be the **next release-please-cut tag push** (probably the v0.1.0 cut from PR #30, if that lands after this PR).

If the attestation step itself fails at release time (e.g. transient sigstore outage), the image push has already succeeded — the release is shipped, but un-attested. That's the desired failure mode: provenance is supplemental, not gating.

## Pre-existing observation (not in scope)

While wiring this up I noticed the existing `Append image digest to release notes` step at the bottom uses `steps.meta.outputs.digest`, but `docker/metadata-action` doesn't emit a `digest` output (only `tags`, `labels`, `version`, `json`, etc.). The shell fallback `${IMAGE_DIGEST:-<digest unavailable>}` likely renders "<digest unavailable>" in release notes today. Per the brief I'm not touching that step in this PR — flagging for a follow-up.

## release-please considerations

No `release-please-config.json` change needed — this is purely a workflow-level addition. Tag-shape is unchanged.

## Verification

- `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/release.yml'))"` — OK
- `uv run ruff check` — passes
- `uv run ruff format --check` — passes
- `uv run ty check` — passes
- `uv run pytest -q --cov-fail-under=80` — 314 passed, coverage 90.91%
- Pinned action SHA verified via `gh api repos/actions/attest-build-provenance/git/refs/tags/v4.1.0` — `a2bbfa25375fe432b6a289bc6b6cd05ecd0c4c32` resolves to the v4.1.0 release commit.
- Action inputs (`subject-name`, `subject-digest`, `push-to-registry`) verified against the action's `action.yml` schema.

## Test plan

- [ ] CI green (lint/type/test/coverage; release-dry-run is unaffected since release.yml itself is touched and dry-run only validates the build step)
- [ ] After v0.1.0 release-please cuts, observe a successful attestation in the workflow run
- [ ] Run `gh attestation verify oci://ghcr.io/verylongorgnamesuchwow/ryzic@<v0.1.0 digest> --repo VeryLongOrgNameSuchWow/ryzic` against the published image; confirm verifier output

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,6 +21,11 @@ on:
 permissions:
   contents: write
   packages: write
+  # Required by actions/attest-build-provenance: id-token mints the OIDC
+  # token sigstore uses to identify the builder; attestations writes the
+  # generated SLSA provenance to the repo's attestations store.
+  id-token: write
+  attestations: write
 
 jobs:
   release:
@@ -58,7 +63,8 @@ jobs:
             type=semver,pattern={{major}}.{{minor}},value=${{ github.event.inputs.tag || github.ref_name }}
             type=raw,value=latest,enable=${{ github.event_name == 'push' }}
 
-      - uses: docker/build-push-action@bcafcacb16a39f128d818304e6c9c0c18556b85f  # v7.1.0
+      - id: build
+        uses: docker/build-push-action@bcafcacb16a39f128d818304e6c9c0c18556b85f  # v7.1.0
         with:
           context: .
           push: true
@@ -66,6 +72,18 @@ jobs:
           labels: ${{ steps.meta.outputs.labels }}
           cache-from: type=gha
           cache-to: type=gha,mode=max
+
+      # Generates SLSA build provenance for the pushed image and uploads it
+      # to the repo's attestation store. Users can verify with:
+      #   gh attestation verify oci://ghcr.io/<repo>@<digest> --repo <repo>
+      # This only succeeds in a real release context (tag push / dispatch
+      # against a tag) because OIDC token issuance for the sigstore signer
+      # requires `id-token: write` from a workflow run authorized to mint one.
+      - uses: actions/attest-build-provenance@a2bbfa25375fe432b6a289bc6b6cd05ecd0c4c32  # v4.1.0
+        with:
+          subject-name: ghcr.io/${{ github.repository }}
+          subject-digest: ${{ steps.build.outputs.digest }}
+          push-to-registry: true
 
       # Release notes are owned by release-please (see release-please.yml).
       # By the time this tag fires, the GitHub Release already exists with


riohno (commented):
## /security-review — PR #38 (build provenance)

**Verdict:** ship as-is
**HIGH-severity findings:** 0

### Findings

None. This PR is a net **reduction** in supply-chain risk: it adds cryptographic provenance from `:tag` → digest → source commit → builder identity for every published GHCR image, with no new attack surface introduced.

### Items checked, no finding

- **Permissions scope** — `release.yml` has a single job (`release`). Workflow-level placement of `id-token: write` + `attestations: write` is functionally equivalent to job-level here; no sibling jobs receive an over-grant.
- **`id-token: write` blast radius** — Audited every step in the job. The OIDC token is consumed only by `actions/attest-build-provenance`. `docker/login-action` and the digest-append step use `secrets.GITHUB_TOKEN` / `GH_TOKEN`, not OIDC. The token is short-lived, audience-scoped to sigstore, and only mintable from this job's runtime context.
- **Action SHA pin** — `actions/attest-build-provenance@a2bbfa25375fe432b6a289bc6b6cd05ecd0c4c32` verified via `gh api repos/actions/attest-build-provenance/git/refs/tags/v4.1.0` → resolves to the same commit. Tag/SHA match confirmed.
- **`subject-name` / `subject-digest`** — `ghcr.io/${{ github.repository }}` uses a trusted GitHub context value (not user-controllable). `${{ steps.build.outputs.digest }}` is the digest of the just-pushed image, emitted by the trusted `docker/build-push-action`. Correct shape; no spoofing vector.
- **`push-to-registry: true`** — Attestation pushed as OCI artifact alongside the image under the existing `packages: write` grant. Publicly verifiable via `gh attestation verify oci://ghcr.io/verylongorgnamesuchwow/ryzic@<digest> --owner VeryLongOrgNameSuchWow` — desired behavior.
- **No secret leakage in SLSA predicate** — `attest-build-provenance` constructs the predicate from GitHub-provided context (workflow ref, commit SHA, runner, builder ID). It does not slurp arbitrary `env:` from the step. The new step defines no `env:` block of its own, and secret-bearing steps (`docker/login-action`, the digest-append shell step) scope their tokens at step level, not job level.
- **Trigger surface** — Workflow only fires on `push: tags: ["v*"]` or `workflow_dispatch` (write-required). PR contributors cannot trigger an OIDC mint here.
- **No template injection** — No new `run:` blocks; only `with:` args to a trusted action, all sourced from trusted contexts.
- **No regression** — Adding `id: build` to the existing `docker/build-push-action` step is purely a label; behavior unchanged. The pre-existing `steps.meta.outputs.digest` bug (issue #39) is unrelated to and not addressed by this PR, per scope.

riohno (commented):
## /simplify — PR #38 (build provenance)

**Verdict:** ✅ already tight (one stylistic nit + a follow-up freebie)

**Estimated trim:** ~3 lines (the stylistic nit), 0 lines of substantive trim

### Wins (in order of value)

1. **Workflow-level vs job-level permissions for `id-token`/`attestations`** — `release.yml:24-28`. Strict least-privilege would scope the two new write perms to the `release` job rather than the workflow top level (matches `scorecard.yml:17,23-27`'s `read-all` + per-job-escalation pattern). With only one job in this workflow the security delta is zero, but it makes the convention legible if a second job (e.g. a future verification job) is ever added without provenance needs. Pure style — no action required if you'd rather match the existing top-level `permissions:` shape used by every other release-family workflow here. (~3 lines moved, not removed.)

2. **Comment trim on the new permissions block** — `release.yml:24-26`. The 3-line comment restates what `actions/attest-build-provenance`'s README already documents. Action name + a link in the PR body would carry the same context. Subjective; the existing repo style is "comment the why, not the what" and the current text is borderline-what. Could compress to a single line: `# Required by actions/attest-build-provenance (OIDC + attestation store).` (saves ~2 lines.)

### Items considered, no win

- `subject-name: ghcr.io/${{ github.repository }}` — **not redundant.** Per `action.yml`, `subject-name` is **required** when using `subject-digest`. Cannot drop.
- `push-to-registry: true` — explicit override of the action's `false` default. Required for the consumer-side `gh attestation verify oci://…` flow promised in the PR body. Keep.
- `id: build` on the build/push step — minimum surface to expose `outputs.digest`. Already as tight as possible.
- The 7-line block comment above the new step — fine. Documents the "verify with…" command and flags the OIDC/release-context constraint, both of which are non-obvious to a reader landing on this file cold.
- Pinned action SHA — explicitly out of scope per brief.

### Bonus (out of scope for this PR — for #39)

The new `id: build` step makes `steps.build.outputs.digest` available in the same job, which is the correct value for the existing `Append image digest to release notes` step (currently reads the non-existent `steps.meta.outputs.digest` and silently falls back to `<digest unavailable>`). One-line follow-up swap: `IMAGE_DIGEST: ${{ steps.meta.outputs.digest }}` → `IMAGE_DIGEST: ${{ steps.build.outputs.digest }}` at `release.yml:75`. Not for this PR — flagged for #39.

---

**Bottom line:** ship as-is. The two nits are stylistic, and the digest-wiring win is correctly deferred to #39 per the PR body.

riohno (commented):
## /review — PR #38 (build provenance)

**Verdict:** ship as-is (one nit + one follow-up suggestion, neither blocking)

### Findings

- **[nit] Permissions could be job-scoped** — `.github/workflows/release.yml:21-28`. `id-token: write` and `attestations: write` are added at workflow level, but `release.yml` has exactly one job (`release`). Workflow-level grants the elevated scopes to any future job added to this file by accident. **Fix:** move the two new permissions (and arguably the existing `contents: write` / `packages: write`) to a `permissions:` block under `jobs.release:`. Strictly tighter; non-blocking because the file is single-job today.

- **[nit] Verifier command not in README** — `README.md` has no `docker pull` or attestation-verify section. The PR body documents `gh attestation verify oci://ghcr.io/.../ryzic@<digest> --repo VeryLongOrgNameSuchWow/ryzic` but operators reading the repo cold will not see it. **Fix (follow-up PR):** add a short "Verifying a release" section to README pointing at the command. Not in scope for this PR.

### Items checked, no finding

- **Action input schema** — `subject-name`, `subject-digest`, `push-to-registry` all present in `action.yml@v4.1.0` with the semantics the PR uses; `push-to-registry: true` correctly requires both `subject-name` (FQ image) and `subject-digest`, both of which are passed. ✅
- **SHA pin matches v4.1.0** — `gh api repos/actions/attest-build-provenance/git/refs/tags/v4.1.0` resolves to `a2bbfa25375fe432b6a289bc6b6cd05ecd0c4c32`. Exact match. ✅
- **Workflow trigger scope** — `release.yml` fires on `push.tags: ["v*"]` and `workflow_dispatch` only. PR CI for release-related changes is in the separate `release-dry-run.yml` (which does not invoke the build/attest path). The PR-body claim "this only runs at release time" is accurate; no risk of an attestation step running in a non-OIDC PR context. ✅
- **Existing softprops digest-append step** — still references `steps.meta.outputs.digest` (the non-existent output captured in #39). This PR does not touch it and does not make it worse: the new `id: build` adds `steps.build.outputs.digest`, but the softprops step still reads the wrong `steps.meta.*` reference. Per scope, leaving it alone is correct — #39 is the fix-up PR. ✅
- **`id: build` collision** — no other step in `release.yml` uses `id: build`; safe to introduce. ✅
- **Failure-mode behavior** — attest step runs *after* `docker/build-push-action`, so a sigstore outage leaves the image pushed but un-attested. That is the documented and correct failure mode for supplemental provenance. ✅
- **release-please / tag shape** — workflow-only change, no `release-please-config.json` impact. ✅

PR #48, opened 2026-05-03
fix(ytdlp): map yt-dlp 'Requested format is not available' to friendly livestream rejection

### LINKED ISSUE #47: /play livestream: friendly 'Live streams aren't supported' message not firing — falls through to raw yt-dlp error
## Summary

The smoke-test doc claims:

> \`/play <livestream_url>\` is rejected with the livestream-not-supported message **before any download starts**.

But on real bring-up of v0.1.0 (smoke-test session), \`/play <livestream_url>\` (test URL: \`youtube.com/watch?v=4RmaQsA9FYs\`) surfaced:

> Could not load that URL. yt-dlp said: ERROR: [youtube] 4RmaQsA9FYs: Requested format is not available. Use --list-formats for a list of available formats

Two failure modes possible:
1. The pre-download \`info.get('is_live')\` (or equivalent) guard in \`src/ryzic/ytdlp.py\` isn't firing.
2. The friendly-error mapping doesn't catch the \`Requested format is not available\` yt-dlp string as a livestream marker.

## Reproduce

\`/play https://www.youtube.com/watch?v=4RmaQsA9FYs\` against any live or recently-live YouTube video.

## Expected

\`Live streams aren't supported. Try a regular video or wait until the stream ends and YouTube publishes the VOD.\` (or whatever the spec'd wording is — sweep \`src/ryzic/ux.py\` for the canonical text).

## Actual

\`Could not load that URL. yt-dlp said: ERROR: ...\` — the raw yt-dlp passthrough. Operators see the URL is rejected (good) but with no actionable message about WHY (bad).

## Severity

**LOW** — operators still get a rejection (no broken playback), just not the friendly variant. **Not a v0.1.0 blocker per smoke-test gating; v0.1.1 patch is fine.**

## Suggested fix

Inspect \`src/ryzic/ytdlp.py\` for the existing \`is_live\` check. If it relies on \`info.get('is_live')\` from a probe call, the probe may itself be failing (the URL might 404 the metadata fetch before \`is_live\` is set). Falling back to string-matching the yt-dlp error message is brittle but may be needed for the recent-but-ended-livestream case.

If the friendly-error map is the gap, add a pattern for \`Requested format is not available\` that maps to the livestream message (with a comment explaining the heuristic).

Pair with a unit test that probes a known live-but-finished URL.

riohno:
## Bug

Smoke-test of v0.1.0 (#47) revealed that `/play <livestream_url>` (e.g. `youtube.com/watch?v=4RmaQsA9FYs`) surfaced a raw yt-dlp passthrough:

> Could not load that URL. yt-dlp said: ERROR: [youtube] 4RmaQsA9FYs: Requested format is not available. Use --list-formats for a list of available formats

instead of the spec'd friendly rejection (`Livestreams are not supported in this version.`).

## Root cause (failure mode 2 from the issue)

YouTube live streams expose only HLS manifests; they do **not** carry the progressive `bestaudio[ext=m4a]/bestaudio[ext=opus]/bestaudio[ext=webm]/bestaudio` formats the wrapper requests. yt-dlp resolves the metadata, then raises `ExtractorError("Requested format is not available", expected=True)` during format negotiation **before** the metadata dict (carrying `is_live: True`) is returned to our wrapper.

Reproduced locally:

```
ERROR: [youtube] 4RmaQsA9FYs: Requested format is not available. Use --list-formats for a list of available formats
```

Without the `format` constraint, `extract_info` *does* return `is_live=True, live_status='is_live'` — but the constraint short-circuits the call. The existing `_is_livestream(info)` post-extract guard never runs for live URLs.

## Fix

Add `Requested format is not available` to `_FRIENDLY_ERRORS` in `src/ryzic/ytdlp.py`, mapped to the existing `_LIVESTREAM_MESSAGE`. Heuristic chosen because:

1. It is the only marker that surfaces for live URLs in our configuration (the metadata-flag path is unreachable).
2. The string is also technically emitted for genuinely-unavailable formats on non-live videos, but our `bestaudio` fallback chain (m4a → opus → webm → bestaudio) always resolves for VODs in practice — so treating it as a livestream signal is safe.
3. The explicit `_is_livestream(info)` guard remains as the primary check for any future code path where format selection succeeds but the flag is set.

The heuristic is documented in-place at the dict definition.

Out of scope (per issue + DO NOT list):
- No changes to `url_validator.py` (runs before yt-dlp).
- No changes to other private/age-restricted/region-locked rejection paths.
- No new dependencies; no rewrite of the error-mapping function.

## Test plan

- [x] Extended `test_resolve_track_maps_known_errors` with the actual smoke-test error string asserting it maps to `Livestreams are not supported in this version.`
- [x] Existing `test_resolve_track_livestream_handling` (parametrized over `is_live`/`live_status`/`was_live`) still passes — covers the metadata-flag path.
- [x] Existing `test_download_rejects_livestream` still passes — covers the download path.
- [x] `uv run pytest -q --cov-fail-under=80` — 315 passed, 90.91% coverage (`ytdlp.py` at 99%)

Closes #47.


riohno (commented):
## /review — PR #48 (livestream rejection mapping)

**Verdict:** ✅ ship as-is

### Findings

**[INFO] Heuristic safety verified against yt-dlp internals** — `src/ryzic/ytdlp.py:48-58` (the new comment block). I traced the exact emission point: `YoutubeDL.py:3058-3060` raises `ExtractorError('Requested format is not available...', expected=True)` from `_select_formats` whenever the format selector returns no candidates. It is not livestream-specific in principle. In practice, however, the wrapper's `bestaudio[ext=m4a]/bestaudio[ext=opus]/bestaudio[ext=webm]/bestaudio` chain terminates in unconstrained `bestaudio`, which matches any audio-bearing format — so the only realistic non-livestream trigger on YouTube would be a video with zero audio-bearing formats (essentially nonexistent). The plausible "false positive" cases I considered (region-lock → already mapped to `Video unavailable`; DRM/Premium-only → uses `raise_no_formats` with a different string; rate-limit → also `raise_no_formats`) emit different error strings and won't reach this branch. The mapping is safe in this configuration. No fix required; the inline comment already documents the tradeoff honestly.

### Items checked, no finding

- **Test specificity** — `tests/test_ytdlp.py:196-200` exercises the actual smoke-session string verbatim (`ERROR: [youtube] 4RmaQsA9FYs: Requested format is not available. Use --list-formats for a list of available formats`), not a generic substring. Strong assertion.
- **`_FRIENDLY_ERRORS` map ordering** — Python dict iteration is insertion-ordered; `next(... if n in detail)` at `ytdlp.py:212` walks in insertion order. The new entry is appended last, so `Sign in to confirm your age` / `Private video` / `Video unavailable` (all more specific) still win when their substrings co-occur. Correct.
- **No regression on the metadata-flag path** — `_is_livestream(info)` and the `resolve_track` / `download` call sites are untouched. Existing parametrized `test_resolve_track_livestream_handling` and `test_download_rejects_livestream` still cover the rarer path where format selection succeeds but `is_live`/`live_status` is set.
- **Constant ordering** — `_LIVESTREAM_MESSAGE` was moved above `_FRIENDLY_ERRORS` so the dict can reference it. Necessary and minimal.
- **CHANGELOG** — `fix:` PR; release-please will render automatically.
- **Scope** — no `url_validator.py` changes, no new dependencies, no rewrite of the error-mapping function. Matches issue #47's "out of scope" list.
- **PR body root-cause writeup** — accurately identifies failure mode (HLS-only manifests vs. progressive-format constraint), and the in-code comment mirrors the reasoning.

PR #51, opened 2026-05-03
feat(ytdlp): add opt-in RYZIC_YOUTUBE_COOKIES_PATH env var for cookies-gated content

### LINKED ISSUE #45: Opt-in YouTube cookies via env var (RYZIC_YOUTUBE_COOKIES_PATH) — enables age-restricted/private content with explicit security disclosure
## Summary

Add an opt-in escape hatch for the deliberate "YouTube cookies disabled" non-feature documented in [README L102](../blob/main/README.md):

> **YouTube cookies are deliberately disabled.** Enabling them lets a deployment fetch age-restricted/private content under your account, which is a security and account-safety risk. There is no env var to enable them in M1; if you need them, fork and review the change carefully first.

The non-feature is correct as a default. But for self-hosters who are running ryzic on hardware they trust + with a YouTube account they accept the credential-leak risk for, an opt-in env var beats a fork-and-edit workflow.

## Proposed shape

\`RYZIC_YOUTUBE_COOKIES_PATH\` env var, **unset by default**. When set, ryzic passes the value as \`cookiefile\` to yt-dlp's options. When unset, current behavior unchanged.

Implementation surface: ~30 LOC in \`src/ryzic/ytdlp.py\` + a \`.env.example\` entry + README disclosure section + tests.

## Required README disclosure (loud — match existing tone)

The new section can't be subtle — it has to spell out the risks:

> **Opt-in: YouTube cookies for age-restricted / private content** (`RYZIC_YOUTUBE_COOKIES_PATH`). When set, ryzic uses the cookies-file at this path on every yt-dlp call.
>
> **Read this before enabling**: anyone who can run a slash command on your bot can now fetch any video your YouTube account can see — including private uploads, age-restricted content, and YouTube Premium-only content. The cookies file effectively grants every \`/play\`-er a session as your account. Compromise of the bot host = compromise of your YouTube account. Do not enable on a server you don't fully control.
>
> Generate cookies via a browser extension (e.g. Get cookies.txt LOCALLY) and store them outside the bot's working directory. Mount as a read-only bind into the container.

## Triggered by

Real smoke test of v0.1.0 (2026-05-03 session): Maddie hit the age-restricted barrier on a mixtape track and confirmed the opt-in is wanted.

## Versioning

Pre-1.0 enhancement → minor bump per `bump-minor-pre-major: true` (already in `release-please-config.json`). Ships as **v0.2.0**. Adds new env var `RYZIC_YOUTUBE_COOKIES_PATH` to the v1.0 stable env-var contract — needs SEMVER.md amendment in the same PR (mirror the kbmcp `KANBANTOOL_READ_ONLY` pattern from kanbantool-mcp#146).

## Out of scope

- Cookie rotation / refresh
- Per-user cookies (multi-tenant; ryzic is single-instance)
- Auto-extracting cookies from a host browser

## Severity

Enhancement, not blocking. v0.1.0 ships without it.

## Tag

`epic` candidate? Borderline — it's one feature in one file but with substantial doc + security framing. Probably OK as a regular issue.

riohno:
## Summary

Adds an opt-in escape hatch (`RYZIC_YOUTUBE_COOKIES_PATH`) for the deliberate "YouTube cookies disabled" non-feature documented in v0.1.0's README. **Unset = current behaviour, byte-for-byte.** Set = `cookiefile` is folded into every yt-dlp call so an operator who accepts the credential-leak risk can play age-restricted, private, and YouTube-Premium-only content without forking the bot.

Pre-1.0 minor bump per `bump-minor-pre-major: true` → ships as **v0.2.0**.

Closes #45.

## What this enables

- Self-hosters who run ryzic on hardware they trust + with a YouTube account they accept the credential-leak risk for can play age-restricted / private / Premium-only content by mounting a `cookies.txt` (Netscape format) and setting `RYZIC_YOUTUBE_COOKIES_PATH` to its in-container path.
- The fork-and-edit workflow the v0.1.0 README pushed people to is no longer required.

## What this risks (and how the disclosure is framed)

The cookies file is **not** a per-user delegation — it grants every `/play`-er a session as the YouTube account that exported the cookies. Compromise of the bot host = compromise of the YouTube account. The README's new "Self-hoster considerations" subsection spells this out verbatim from the issue's required-disclosure block (terse, names the credential-leak risk, says "do not enable on a server you don't fully control"). The `.env.example` carries a matching WARNING block and is shipped commented-out, and `bot.py` logs a loud `WARNING` on every startup when the env var is non-empty so an operator can't quietly forget the cookies are live.

## What changes

- `src/ryzic/config.py` — new `Config.youtube_cookies_path: Path | None` parsed from the env var. Empty string and unset both resolve to `None`.
- `src/ryzic/ytdlp.py` — module-level singleton + `set_cookies_path()` setter (symmetric to `audio_cache.set_audio_cache`). `_base_opts()` adds `cookiefile` only when set; absent key preserves yt-dlp's cookie-less default. `cookiesfrombrowser` stays unconditionally disabled. yt-dlp owns format validation; we don't duplicate.
- `src/ryzic/bot.py` — reads `cfg.youtube_cookies_path` at startup, logs a WARNING when active, installs into the singleton.
- `.env.example` — commented-out section with the WARNING-prefixed disclosure.
- `README.md` — pre-existing "deliberately disabled" bullet now points at the new opt-in subsection. New `### Opt-in: YouTube cookies for age-restricted / private content` subsection with the issue's required wording + a compose-bind-mount example. New row in the configuration table.
- `SEMVER.md` — v1.0+ stable env-var contract enumerated explicitly (every name, including `RYZIC_YOUTUBE_COOKIES_PATH`), mirroring the kanbantool-mcp pattern of committing env-var names to the stable surface in the same PR that adds them.

Compose files are intentionally untouched: the opt-in pattern means operators add the env line + bind mount to their own `compose.yaml` (the README example shows the exact shape).

## Tests

- `tests/test_ytdlp.py` — assertion swap on the existing security-critical opts test (`cookiefile` is now absent by default rather than `None`-valued) plus 3 new cases covering unset, set, and clear-after-set behaviour. New autouse fixture restores the singleton between tests so the writer can't leak under `pytest-randomly`.
- `tests/test_config.py` (new) — 3 cases covering unset, empty-string, and set env-var → `Path | None` mapping.

No integration test for actually fetching age-restricted content — out of scope and unfair to YouTube.

## Test plan

- [x] `uv run pytest -q --cov-fail-under=80` — 321 passed (was 314); coverage 90.77%
- [ ] CI green
- [ ] Manual smoke (post-merge, against a real account): set the env var, `/play` an age-restricted URL, confirm the previously-friendly rejection now resolves to playback. Without the env var set, behaviour identical to v0.1.0.


riohno:
## Review of PR #51 — `feat(ytdlp): add opt-in RYZIC_YOUTUBE_COOKIES_PATH`

**Verdict: Approve with nits.** No blockers. The feature is correctly scoped, the security disclosure matches the issue's required wording verbatim, the security-critical "no `cookiefile` key when unset" assertion is in place, the singleton mirrors the existing `audio_cache` convention, and CI is green (321 passing, coverage 91%, ruff/format/ty clean). I also re-ran the suite under five `pytest-randomly` seeds against the worktree at `44ad9f3` — autouse cookies-path-restore fixture holds; no order-dependent leaks.

Findings below are all non-blocking. None of them need to land in this PR.

---

### 1. nit (suggestion, not actionable here): PR is `BEHIND` `main`

`gh pr view 51 --json mergeStateStatus` reports `BEHIND` — release `0.1.1` (`cc68785`) landed on `main` after this branch was cut. Strict branch protection requires an update-branch before `gh pr merge` will go through. Not a code issue; just a heads-up that a `gh pr update-branch 51` (or a rebase + force-push) is needed before merge.

### 2. nit (consider but skip if it's churn): startup file-existence check

`bot.py` accepts the configured `cookies_path` and installs it without verifying the file exists. A typo'd path (`/etc/ryzic/youtbe-cookies.txt`) gets discovered only on the first `/play`, surfacing as a yt-dlp `IOError` mapped through the generic `Could not load that URL. yt-dlp said: \`...\`` path. A startup `if not cfg.youtube_cookies_path.is_file(): raise ConfigError(...)` would fail-fast and match the "validation happens at startup so a missing or malformed value fails the process immediately" docstring on `config.py`.

Counter-argument and reason this is **just a suggestion**, not a request: the project deliberately doesn't validate `cache_dir` until `audio_cache.open()` runs either. Adding this check for cookies but not cache would be inconsistent. Either skip or follow-up holistically.

### 3. nit (skip for this PR): smoke-test checklist not amended

`docs/manual-smoke-tests.md` doesn't carry a step exercising `RYZIC_YOUTUBE_COOKIES_PATH`. A pre-release tester running the existing checklist will not exercise the opt-in path, so any regression in the cookies plumbing would slip past the manual smoke gate. The PR description's unchecked "Manual smoke (post-merge)" item covers the immediate v0.2.0 cut, but adding a permanent step under `## Error handling` (or a new `## Opt-in YouTube cookies` subsection) — guarded by "skip if `RYZIC_YOUTUBE_COOKIES_PATH` is unset" — would lock the verification in for future releases.

Reasonable to defer to a follow-up issue rather than expand this PR's scope.

### 4. suggestion (defer): no test asserts the startup WARNING fires

`bot.py:main()` is at 22% coverage (pre-existing — the lifecycle entrypoint isn't unit-tested), and the new WARNING-on-opt-in branch lives inside it. The "loud on every restart" guarantee in the README + .env.example isn't backed by an automated assertion. Given the existing convention this isn't a regression and isn't a blocker, but a small `caplog`-based test that calls a refactored helper (e.g. extracting `_install_cookies(cfg)` from `main`) would tighten the contract. Defer if it'd require restructuring `main()` more than feels worth it.

### 5. nit (skip): minor doc consistency

Pre-existing, not introduced here, but worth noting alongside the SEMVER amendment: `.env.example` line 20 still enumerates log levels as `DEBUG, INFO, WARNING, ERROR.` (missing `CRITICAL`), while `config.py` accepts `CRITICAL/ERROR/WARNING/INFO/DEBUG` and the README table says the same. Out of scope for this PR, but worth a 1-line `docs:` follow-up given SEMVER.md is now committing the env-var contract verbatim.

---

### What I checked and confirmed

- **`cookiefile` plumbing**: env var → `Config.youtube_cookies_path: Path | None` → `ytdlp.set_cookies_path()` → module global `_COOKIES_PATH` → conditional `opts["cookiefile"] = str(...)` in `_base_opts()`. Unset path is the absent-key form, **not** `cookiefile=None`, which is exactly what the README's "preserves the cookie-less default" wording promises. Verified by re-reading `_base_opts()` and the security-critical assertion (`assert "cookiefile" not in opts`).
- **Singleton symmetry**: `set_cookies_path` matches the shape of `set_audio_cache` (module-level `_COOKIES_PATH`, setter writes the global, no getter because `_base_opts` is the sole reader). KISS-compliant — passing the path through every yt-dlp callsite would mean threading config through `resolve_track`, `resolve_playlist`, `download`, plus their `audio_cache` callers. Singleton is the right call here.
- **README anchor**: `#opt-in-youtube-cookies-for-age-restricted--private-content` correctly resolves to the new heading per GitHub's slug algorithm (drop punctuation, lowercase, replace spaces with `-`; the `/` becomes a double-dash because the surrounding spaces collapse onto each other).
- **Test coverage**: unset / set / clear-after-set / empty-string / non-`None`-but-cleared all covered. The autouse `_restore_cookies_path` fixture is present and works under randomized order.
- **SEMVER.md amendment**: the "v1.0+ stable surfaces" env-var enumeration is comprehensive and correctly includes `RYZIC_YOUTUBE_COOKIES_PATH` alongside the existing eight names.
- **Security disclosure**: the README's "Self-hoster considerations" subsection and `.env.example` WARNING block both spell out "compromise of the bot host = compromise of your YouTube account", "do not enable on a server you don't fully control", and the credential-leak framing — exactly matching the issue's required disclosure wording.
- **Comment quality**: comments are WHY-only (`SECURITY:` framing, the symmetry-to-`set_audio_cache` justification, the autouse-fixture rationale). No WHAT-narrating prose.
- **No new dependencies**, no premature abstractions, no DRY violations.

Approve with nits — happy to revisit if any of the suggestions feel worth folding in pre-merge.

riohno:
## Security review — PR #51 (`feat/youtube-cookies-opt-in`, head `44ad9f3`)

**Verdict: Findings present — 1 HIGH blocker, 1 MEDIUM blocker, plus 4 LOW improvements.**

The architecture is sound (single-setter singleton, opt-in default-absent key in opts, no subprocesses, browser-cookies extraction stays unconditionally disabled, scrub regex catches the cookies path in user-visible errors, env read once at startup, no log line ever contains cookie *contents*). The disclosure copy is strong on the credential-leak risk. The blockers below are operational footguns rather than direct credential-leak vectors, but both will hit operators on first use.

---

### Finding 1 — HIGH (confidence 10/10): the README's recommended `:ro` cookies mount makes every `/play` fail

**Severity:** HIGH (blocker — documented happy path is broken)
**Confidence:** 10/10

`yt-dlp`'s `YoutubeDL.__exit__` unconditionally calls `save_cookies()` whenever `cookiefile` is set, which `open(file, 'w')`s the cookies path. With the README's recommended read-only bind mount, this raises `PermissionError` on context exit *after* a successful extraction. The exception propagates out of `_sync_extract` → `_extract`'s bare `except Exception` → user sees `FetchFailed("internal error: PermissionError")` and the bot logs a full traceback at ERROR. Reproduced end-to-end against the PR branch with the cookie file at mode `0o400` and `extract_info` mocked to a successful result:

```
File ".../yt_dlp/cookies.py", line 1305, in open
    with open(file, 'w' if write else 'r', encoding='utf-8') as f:
PermissionError: [Errno 13] Permission denied: '/tmp/...cookies.txt'
FAIL: FetchFailed: internal error: PermissionError
```

This means an operator who follows the README literally — and the README *explicitly* tells them to mount with `:ro` ("Mount as a read-only bind into the container") — gets a bot that can't play *any* track once cookies are enabled, not just gated content. The bot will only work if the operator deviates from the docs and mounts read-write, which then introduces a different problem: yt-dlp silently rewrites the cookies file with refreshed session tokens on every call, requiring the file to be writable by the container's bot user (UID needs to match the host file owner).

**Remediation (pick one):**

1. **Wrap the cookies file in a private writable copy at startup.** In `bot.py` after `set_cookies_path`, copy the operator-supplied file into `cfg.cache_dir / "youtube-cookies.txt"` (mode 0o600, bot-owned) and pass *that* path to yt-dlp. The original `:ro` mount stays the source of truth; yt-dlp gets a writable scratch copy it owns. Trade-off: refreshed session tokens stay inside the container and are lost on restart, which is fine for short-lived bot processes but means the operator periodically re-exports the cookies file (already implicit since browser cookies expire).
2. **Pass an in-memory file-like instead of a path.** `cookiefile` accepts a text stream; opening the operator's file `'r'` once at startup and handing yt-dlp the stream object means `cookies.py:open()` takes the non-`is_path_like` branch and `f.truncate(0)` will fail loudly on a read-only stream — but the failure surfaces as a no-op write, not a `PermissionError`. Less invasive but requires re-reading the file if it rotates on host.
3. **Document `:rw` instead and add a `chown` instruction.** Honest-but-uglier: tell operators the file *must* be writable by the bot's UID, drop the `:ro`, accept that yt-dlp will overwrite the file. This is what other yt-dlp consumers tend to do, but it weakens the "ryzic can't tamper with the source cookies" property.

Option 1 is recommended — it preserves the `:ro` source posture, contains yt-dlp's writes inside the bot's existing cache volume (already private to the bot), and matches the bot's "single-instance per cache dir" assumption. Add a regression test that calls `resolve_track` with the singleton pointing at a 0o400 file and asserts no `FetchFailed("internal error: …")` surfaces.

---

### Finding 2 — MEDIUM (confidence 8/10): silent no-op on a typo'd cookies path

**Severity:** MEDIUM (blocker per the threat model — explicit footgun the PR was asked to prevent)
**Confidence:** 8/10

`config._parse_optional_path` only checks `raw is None or raw == ""`; it does not stat the path. If the operator sets `RYZIC_YOUTUBE_COOKIES_PATH=/etc/ryzic/youtibe-cookies.txt` (typo), `bot.py` logs the loud WARNING `"YouTube cookies enabled from /etc/ryzic/youtibe-cookies.txt"` — so the operator believes cookies are live — but yt-dlp's `load_cookies` checks `os.access(cookie_file, os.R_OK)` and silently skips the load when the file doesn't exist. Age-restricted URLs then surface the friendly "That video is age-restricted and can't be played" message, with no signal pointing back at the misconfigured cookies path.

Threat-model question #6 explicitly called this out: the operator thinks cookies are enabled but they're not, and only finds out from a confusing rejection downstream.

**Remediation:** in `config._parse_optional_path` (or a new validator called from `load()`), require the path to exist and be a regular file when `RYZIC_YOUTUBE_COOKIES_PATH` is set. Raise `ConfigError("RYZIC_YOUTUBE_COOKIES_PATH points at a path that does not exist or is not a regular file: …")` so the bot fails fast at startup. This matches the "fail-fast at startup" doctrine in `config.py`'s docstring and the existing pattern for `_parse_log_level` / `_parse_positive_int`. Optionally also check `os.R_OK` for the bot's UID to catch permission misconfigurations the same way.

---

### Finding 3 — LOW (confidence 8/10): `youtube_cookies_path` is not `field(repr=False)` — divergence from the bot-token / lavalink-password convention

**Severity:** LOW
**Confidence:** 8/10

`Config.discord_bot_token` and `Config.lavalink_password` are `field(repr=False)` to keep secrets out of any incidental `repr(cfg)`. `youtube_cookies_path` is not. The path itself is not a credential, but it points at a file containing session tokens, and the existing `tests/test_imports.py::test_config_repr_hides_secrets` enshrines the convention. Future work that prints `repr(cfg)` (a debug log, an error context, a config-dump command) would expose the cookies path.

**Remediation:** mark the field `field(repr=False, default=None)` and extend `test_config_repr_hides_secrets` to include the cookies path. Pure consistency / defense-in-depth.

---

### Finding 4 — LOW (confidence 6/10): cookies-path leakage to the bot's own logs via yt-dlp's stderr writes

**Severity:** LOW
**Confidence:** 6/10

When the cookies file is malformed, yt-dlp prints `ERROR: '/etc/ryzic/youtube-cookies.txt' does not look like a Netscape format cookies file` directly to stderr *despite* the wrapper setting `quiet=True` and `no_warnings=True`. `_scrub` masks the path on the user-facing `FetchFailed` string (good), but the host's stderr/log capture still sees the raw path. The path itself is not secret — the operator put it there — but it's noise that violates the wrapper's "we own the error pipeline" intent.

**Remediation:** either accept the leak and document that the operator's own logs may surface the cookies path, or wrap the `with YoutubeDL(opts):` block in a `contextlib.redirect_stderr(io.StringIO())` and emit only the wrapper's own `_log.exception` line. The latter risks suppressing useful diagnostics; given the path is operator-controlled and not a credential, "accept and document" is the lower-cost fix.

---

### Finding 5 — LOW (confidence 7/10): disclosure misses the post-compromise remediation step

**Severity:** LOW
**Confidence:** 7/10

The README's "Self-hoster considerations → Opt-in" subsection covers (a) credential-leak risk and (b) "do not enable on a server you don't fully control" — both well. It's missing (c): the action the operator should take if they *do* suspect bot-host compromise. The threat model question #5 called this out specifically.

**Remediation:** append one sentence: *"If you suspect the bot host was ever compromised while cookies were active, sign out all sessions on the YouTube account (Google → Manage your Google Account → Security → Your devices) and re-export a fresh cookies file."* Roughly the same shape as the existing 3-bullet copy; no other doc surface affected.

---

### Finding 6 — LOW (confidence 6/10): cookie file ownership and host-side perms not addressed in docs

**Severity:** LOW
**Confidence:** 6/10

The README's compose snippet shows `volumes: - /path/on/host/youtube-cookies.txt:/etc/ryzic/youtube-cookies.txt:ro` but says nothing about host-side perms. A cookies file at `0o644` owned by the operator's user would be world-readable on the host — readable by any user logged into the host (a typical multi-user VPS scenario the README's "do not enable on a server you don't fully control" line gestures at, but doesn't quite cover). Similarly there's no guidance on storing the file outside the deploy directory (a `cp cookies.txt ./` followed by a `git add .` is a plausible accident path).

**Remediation:** in the README's opt-in subsection, append: *"Set the host-side file to `chmod 0o600` and store it outside the directory containing your `compose.yaml` and `.env` to keep it from being accidentally committed or `tar`-bundled with the deployment."* No code change required.

---

### Out of scope for this review but noticed

The PR branch reverts `pyproject.toml`, `.release-please-manifest.json`, and `CHANGELOG.md` from `0.1.1` back to `0.1.0` (visible in the diff against `origin/main`). The branch was likely created from a SHA before #49's release-please merge. This isn't a security issue, but the merge will fight release-please at minimum, and at worst clobber the v0.1.1 release notes from CHANGELOG. Rebase onto current `main` before merging.

---

### Architecture aspects that look correct

For the record, the threat-model items the PR handles cleanly:

- **#1 path-traversal / arbitrary-file-read:** yt-dlp loads cookies via `MozillaCookieJar`, which only reads the file as text and parses Netscape format; no shell expansion beyond `expand_path` (which only honors `~` and `$VAR` from the bot's already-scoped env). Not a meaningful attack surface here.
- **#2 cookie-content log leakage:** wrapper never logs `opts`, never logs cookie contents. The path is logged at WARNING by intent. yt-dlp's `MozillaCookieJar` doesn't print values either.
- **#4 default safe:** `cookiefile` is *absent* from the opts dict by default rather than `None`-valued — covered by the assertion swap in `test_base_opts_security_critical_settings` and the dedicated `test_base_opts_no_cookiefile_when_unset`. Empty string and unset both resolve to `None` in `_parse_optional_path` — covered by `test_youtube_cookies_path_unset_resolves_to_none` and `test_youtube_cookies_path_empty_resolves_to_none`.
- **#7 symlink / TOCTOU:** path is captured eagerly into a frozen `Config`, then handed to yt-dlp which `expand_path`s it once. No `Path.resolve` race window the wrapper introduces. The host filesystem races (operator-rotates-cookies-mid-call) are out of scope.
- **#8 process-env exposure:** env read once at startup (`config.load()`), held in the frozen `Config`, never re-read. No subprocesses (yt-dlp embedded API only). No path for the env value to leak to a child process.
- **`cookiesfrombrowser` stays unconditionally disabled** with a clear comment explaining why — good narrowing of the blast radius.
- **Test isolation:** the autouse `_restore_cookies_path` fixture correctly contains the singleton mutation, so `pytest-randomly` can't bleed it.

riohno:
## Code review — fix delta `44ad9f3..c4aa408`

**Verdict: Approve, 0 blockers.**

Clean fix. The scratch-copy approach preserves the README's `:ro` mount posture, the validation closes the silent-no-op gap from security review §2, and the test additions are well-targeted. Local run reproduces the implementer's claims: 334 passed, coverage 91.09%, ruff/format/ty all green.

Reviewing only the delta on top of `44ad9f3` (original feature review's nits not relisted).

### Findings

1. **(suggestion) `_parse_existing_file` reads `os.environ.get(name)` twice.** Lines 101 and 104 of `config.py`: `_parse_optional_path` already reads the env var, then `_parse_existing_file` reads it again to get `raw` for the error message. Functionally fine (env doesn't mutate during `load()`), but the second lookup is the kind of small redundancy a future reader will pause on. Optional refactor: have `_parse_optional_path` return both the `Path` and the `raw` string (tuple), or have `_parse_existing_file` read `raw` once at the top and pass the parsed `Path` through manually. Not worth holding the PR.

2. **(suggestion) Symlink-as-source not explicitly tested.** `Path.is_file()` follows symlinks (verified locally), so an operator who symlinks their cookies file into the documented mount path passes validation — that's the correct user-friendly behavior. Adding a one-line test pinning that contract would prevent a future "tighten validation" change from accidentally breaking symlink-resolved paths. Skip if you'd rather keep the test surface minimal.

3. **(nit) Module-constant docstring is borderline WHAT-narration.** `bot.py` lines 16-18: "Filename of the writable scratch copy ... so tests can refer to the same name without hard-coding a literal in two places." The "why a constant" justification is real, but the lead sentence reads like a docstring describing the value. Trimmable to one line if you tighten the file later; not worth a separate commit.

### Per-criterion notes

- **Helper ordering / `cache_dir` creation / log levels**: ordering in `main()` is correct (`basicConfig` → `_install_youtube_cookies` → `_bootstrap_audio_cache`). `mkdir(parents=True, exist_ok=True)` reliably creates the cache dir on first startup. WARNING for the security-relevant opt-in (so operators see it on every restart even at default log levels) + INFO for the scratch-path transparency note is the right split.
- **`_parse_existing_file` design**: layering on top of `_parse_optional_path` keeps the generic helper generic; SRP is clean. Error messages are actionable and consistent with `_parse_log_level` / `_parse_positive_int` style (named env var, offending value, fix hint).
- **Test coverage**: 10 tests in `test_bot.py` (the brief said 9, the diff has 10 — undercount, not over) + 3 in `test_config.py`. Each exercises a distinct behavior: defensive setter clear, SRP separation when unset, scratch path vs source, content match, mode 0o600, WARNING source-path content, INFO scratch-path content, rotation rebuild, source-byte preservation, cache_dir auto-create, missing-file rejection, directory rejection, unreadable rejection. No padding. Autouse `_restore_cookies_path` mirrors `tests/test_ytdlp.py` correctly — `original` is captured before `yield`, so pre-pollution in `test_install_cookies_unset_clears_setter` doesn't bleed. `pytest-randomly`-safe.
- **Comment hygiene**: all new comments earn their place — they explain WHY (yt-dlp's `save_cookies()` behavior, fresh-rebuild for rotation, defense-in-depth chmod, WARNING-vs-INFO rationale, why `repr=False` on the cookies path). One borderline (finding 3 above).
- **DRY**: `_install_youtube_cookies` and `audio_cache.AudioCache.open()` both call `mkdir(parents=True, exist_ok=True)` on `cache_dir`. This is appropriate SRP — each component owns its own preconditions, the helper runs synchronously before the async cache bootstrap, and a one-line idempotent mkdir doesn't warrant extracting a shared helper.
- **README delta**: post-compromise sentence sits well at the end of the existing "Read this before enabling" warning — completes the threat→remediation pattern. Scratch-copy disclosure (after the compose example) is short and operator-relevant: explains why their `:ro` mount works and matters for backup/snapshot policy. Not noise.
- **`uv.lock`**: only diff is `version = "0.0.0"` → `version = "0.1.1"` on the `name = "ryzic"` package block. Benign reconciliation after the v0.1.1 release-please merge from main, no real dependency change.
- **Out of scope per Maddie**: not asking the implementer to address LOW #4 (yt-dlp stderr) or LOW #6 (host perms).

riohno:
## Security re-review — PR #51 (`feat/youtube-cookies-opt-in`, head `c4aa408`)

**Verdict: Both blockers closed. One new LOW finding (stale-credential lifecycle); LOW-6 from the original review remains open.**

I verified each fix end-to-end against real yt-dlp (not just the PR's mocks) and probed the new attack surface introduced by the scratch-copy logic. The architecture choice (Option 1) is correctly executed and the regression test coverage is solid. The one new finding is a credential-lifecycle issue that's worth flagging but doesn't gate the merge.

---

### Original Finding 1 — HIGH (`44ad9f3`): `:ro` mount makes every `/play` fail — **CLOSED**

**Verification:** end-to-end repro against the new code with real yt-dlp:

- Created a 0o400 source cookies file (matches the README's `:ro` mode).
- Set `RYZIC_YOUTUBE_COOKIES_PATH` at it; ran `_install_youtube_cookies(cfg)`.
- Confirmed the singleton ends up holding the **scratch** path (`cache_dir/youtube-cookies.txt`), not the source path — so `_base_opts()["cookiefile"]` points at the writable copy.
- Entered/exited a real `YoutubeDL(opts)` context manager with `extract_info` mocked. Previously this would `PermissionError` inside `cookies.py:1305`. Now it exits cleanly — `save_cookies()` writes to the 0o600 scratch file successfully.
- Source file is byte-identical after the round trip; source mode is still 0o400 (operator's `:ro` posture preserved).

Walking the checklist:

1. **Helper runs before set_cookies_path on source path?** Yes — `bot.main()` calls `_install_youtube_cookies(cfg)` at line 115, which only invokes `ytdlp.set_cookies_path(scratch_path)` (line 86) after the copy succeeds. The singleton can never hold the source path.
2. **Fresh copy each startup?** Yes — `shutil.copyfile` overwrites unconditionally (verified by the regression test `test_install_cookies_rebuilds_scratch_on_each_call` and the comment at `bot.py:64-66`). Rotated cookies take effect on restart.
3. **Scratch lands in `cfg.cache_dir`?** Yes — `cache_dir / _COOKIES_SCRATCH_FILENAME` (`bot.py:62`). Same volume the audio cache uses.
4. **Mode 0o600?** Yes — explicit `scratch_path.chmod(0o600)` at `bot.py:71`, with a comment noting it's set regardless of the operator's umask. Verified by `test_install_cookies_scratch_copy_is_mode_0o600` against a wide-open 0o644 source.
5. **`:ro` failure simulated against new code?** Yes — both via the implementer's `test_install_cookies_does_not_modify_source` (mocks the yt-dlp write) and via my own e2e probe with a real `YoutubeDL` context manager and a 0o400 source. PermissionError is gone.
6. **WARNING references source path, not scratch?** Yes — the loud WARNING at `bot.py:76-80` interpolates `cfg.youtube_cookies_path` (the operator-recognized path); the scratch path is logged separately at INFO. Locked in by `test_install_cookies_warning_references_source_path_not_scratch` and `test_install_cookies_info_logs_scratch_path`.
7. **Race conditions in copy?** `shutil.copyfile` is non-atomic; a crash mid-copy leaves a truncated scratch. Acceptable per the original review note — next startup re-copies fresh, and yt-dlp would parse-reject a partial Netscape file (effectively no cookies in flight) rather than send a malformed credential. Not worth a temp-file-+-rename here.

---

### Original Finding 2 — MEDIUM (`44ad9f3`): silent no-op on a typo'd cookies path — **CLOSED**

**Verification:** new validator `_parse_existing_file` (`config.py:91-115`) replaces `_parse_optional_path` for the cookies env var (`config.py:128`). Walking the checklist:

1. **Catches the three failure modes?** Yes — `path.is_file()` rejects both missing-file and directory cases; `os.access(path, os.R_OK)` catches permission misconfigurations. Each gets a distinct `ConfigError` with an actionable message.
2. **Raises at `config.load()`?** Yes — `_parse_existing_file` is called from `load()` so the bot fails fast at startup.
3. **Error message actionable?** Yes — both messages name the env var (`RYZIC_YOUTUBE_COOKIES_PATH`), the bad raw value, and what to check. The unreadable-file message specifically suggests `chmod 0o600` owned by the bot UID.
4. **Pattern consistent with `_parse_log_level` / `_parse_positive_int`?** Yes — same shape (`_parse_<x>(name)`, `ConfigError(f"{name}: ...")`), same fail-fast docstring intent.
5. **Tests?** Three new tests in `tests/test_config.py` cover all three modes: `test_youtube_cookies_path_missing_file_raises`, `test_youtube_cookies_path_directory_raises`, `test_youtube_cookies_path_unreadable_raises`. All pass under `uv run pytest -q`.

---

### Original LOW findings — status

- **LOW-3** (`field(repr=False)` on `youtube_cookies_path`): **CLOSED.** Field is now `field(repr=False, default=None)` (`config.py:34`) with a comment explaining why; `test_config_repr_hides_secrets` extended with the cookies path.
- **LOW-4** (yt-dlp stderr leaking the cookies path): **NOT addressed.** Was an "accept and document" suggestion; no behaviour change required and operator-controlled paths aren't credentials. Fine to defer.
- **LOW-5** (post-compromise remediation copy): **CLOSED.** README §"Opt-in: YouTube cookies" now appends the "sign out all sessions" guidance.
- **LOW-6** (host-side perms guidance): **NOT addressed.** README still doesn't tell operators to `chmod 0o600` the host-side cookies file or store it outside the deploy dir. Worth doing in a follow-up PR (one-sentence README addition); not a blocker.

---

### New finding — LOW (confidence 8/10): stale credential file lingers in `cache_dir` after operator disables cookies

**Severity:** LOW
**Confidence:** 8/10

`_install_youtube_cookies` only writes the scratch when `cfg.youtube_cookies_path is not None`; it does not unlink an existing scratch when the operator unsets `RYZIC_YOUTUBE_COOKIES_PATH`. Verified in a temp dir:

- Boot 1 with cookies set → `cache_dir/youtube-cookies.txt` exists at 0o600 with the cookie contents.
- Boot 2 with `RYZIC_YOUTUBE_COOKIES_PATH` unset → singleton is correctly cleared via `set_cookies_path(None)` so yt-dlp doesn't load it, **but the scratch file remains on disk** with the same cookie contents.

Operationally: an operator who decides to disable the cookies opt-in (because they realized the security implications, or because the YouTube account got compromised) and restarts the bot will believe the credential has been removed from the bot's filesystem. It hasn't. It sits in the cache volume (which is bot-owned, 0o600, so the leak surface is small — anyone who can read `cache_dir` already has the bot UID). The risk is the operator's mental model: they think they've cleaned up; they haven't.

**Remediation:** in the `cfg.youtube_cookies_path is None` branch of `_install_youtube_cookies`, also unlink `cfg.cache_dir / _COOKIES_SCRATCH_FILENAME` (with `missing_ok=True`). One line, plus a regression test that asserts the scratch is gone after a None-config call when a stale scratch existed beforehand.

Not a merge blocker — the threat is "operator's mental model diverges from disk state," not a credential-leak vector — but it's the kind of footgun the original PR was specifically scoped to prevent and the fix is trivial.

---

### Other concerns probed and dismissed

- **Symlink follow:** `shutil.copyfile` follows symlinks by default (verified empirically — copied file contents land in the scratch when the source is a symlink). The threat model in the prompt addressed this: a malicious local user swapping the operator's source path for a symlink to `/etc/passwd` would have its contents copied into the 0o600 scratch, but yt-dlp parses Netscape format and rejects non-cookie content, so the copied bytes never leave the bot. Operator-side, if the cookies path is a deliberate symlink to a real cookies file elsewhere on the host, that's the operator's choice. Benign.
- **Cache directory permissions:** `_install_youtube_cookies` calls `cfg.cache_dir.mkdir(parents=True, exist_ok=True)` without setting mode — relies on the umask. On a default umask (0o022) that lands at 0o755, world-readable. The 0o600 scratch file inside is fine on its own (its contents aren't readable to a stranger), but a world-readable cache dir means anyone can enumerate which bots have cookies enabled (the file's existence is the signal). Already acknowledged downstream — `ytdlp.py:313` says "the cache directory's permissions (0o700, bot-owned) are the deploy-time mitigation," which is operationally correct (Docker volumes default to root-owned 0o755 but the running container has the bot UID). Not a code change for this PR.
- **TOCTOU between validation and copy:** `_parse_existing_file` validates at `config.load()` time; `_install_youtube_cookies` copies seconds later. A malicious local user could swap the source file in between. Out of scope per the original review's threat model (#7 — single-user trust on the host); the README's "do not enable on a server you don't fully control" line covers it.
- **Scratch path leakage in logs:** the INFO line includes the scratch path. The path itself is not a credential (it's a deterministic location inside the bot's cache); it serves as transparency that the scratch-copy step ran. Deliberate per the implementer's comment.
- **shutil.copyfile failure handling:** verified that a `PermissionError` raised from `copyfile` propagates out of `_install_youtube_cookies` (no try/except) without ever reaching `set_cookies_path`. The bot fails-fast at startup with the exception trace; the singleton stays at its prior value (None on a fresh boot). Correct fail-closed behaviour.
- **Rebase concern from the original review's "out of scope" section:** the version-file revert is gone — `git diff origin/main..origin/feat/youtube-cookies-opt-in` shows no diff in `pyproject.toml`, `CHANGELOG.md`, or `.release-please-manifest.json`. Branch was successfully merged forward.

---

### Test suite + tooling

- `uv run pytest -q tests/test_bot.py tests/test_config.py tests/test_imports.py tests/test_ytdlp.py` — 71 passed. The new `tests/test_bot.py` (10 cases) and the 3 new `tests/test_config.py` cases all pass.
- `uv run ruff check` — clean.
- `uv run ruff format --check` — 39 files already formatted.
- `uv run ty check` — clean.

---

### Summary

Both `44ad9f3` blockers are genuinely closed with end-to-end evidence (not just mock-level). The scratch-copy approach is correctly executed (mode-0o600, fresh-each-boot, source untouched, fail-fast on copy errors) and the path validator gives operators an actionable error at startup. One new LOW (stale scratch after disable) and one carry-over LOW (host-side perms doc) remain — neither blocks the merge. **Ship it.**

PR #77, opened 2026-05-03
feat(config): add RYZIC_AUTOLEAVE_SECONDS for configurable queue-end auto-leave

### LINKED ISSUE #62: RYZIC_AUTOLEAVE_SECONDS env var (configurable auto-leave timer)
## Context
Sourced from the UX/QoL audit (overnight 2026-05-02). 17-item roadmap with proposal-writer + principles-guardian + competitor-survey + user-pain-mining workstreams.

## Proposed shape
Add `RYZIC_AUTOLEAVE_SECONDS` env var to control the post-queue-empty auto-leave timer. Default 300 (preserves current behaviour). Setting 0 disables the auto-leave entirely (24/7 ambient playback use case).

## Hard lines / watch-outs
- Default value MUST be 300s to preserve existing behaviour for all current operators.
- `0` disables; document explicitly.
- Asymmetric outcome: the empty-channel timer in #7 stays HARDCODED — different evidence, no env var. Don't generalize the env-var pattern across both timers.

## Decision (2026-05-03)
SHIP at v0.3.0.
~15 LOC, low-risk, ops-focused; clusters cleanly with the v0.3.0 small-ops batch.

## Source
Morning UX/QoL audit team: `competitor-survey`, `user-pain-mining`, `principles-guardian`, `proposal-writer` + lead.

riohno:
## Summary

Adds `RYZIC_AUTOLEAVE_SECONDS` env var controlling the post-queue auto-leave timer (closes #62, item #5 from the morning UX/QoL audit roadmap).

- Default `300` preserves the historical 5-minute window — no behaviour change for existing operators.
- `0` disables the timer entirely; the bot stays connected indefinitely after the queue empties (24/7 ambient music server use case).
- Validation: integer `>= 0`; non-integer / negative values raise `ConfigError` with an actionable message at startup.

## Asymmetric outcome (intentional)

This env var only covers the **queue-end** auto-leave timer. The **empty-channel** timer (P4 / audit item #7, scheduled for v0.7.0) stays hardcoded for now — different evidence, different decision, NOT cargo-cult symmetry across the two timers. Reviewers should not flag the absence of an empty-channel env var as a missing surface; ship it independently when the field signal arrives.

## Surfaces touched

- `src/ryzic/config.py` — new `auto_leave_seconds` field on `Config`; new `_parse_non_negative_int` helper (allows `0` where `_parse_positive_int` rejects it).
- `src/ryzic/lavalink_glue.py` — `_start_auto_leave` skips scheduling when the value is `0` (a 0-second sleep would defeat the operator's intent); `_auto_leave` passes the configured seconds into the message; `_format_idle_duration` renders the natural unit so the default still shows "5 minutes" not "300 seconds".
- `.env.example`, `README.md`, `SEMVER.md`, `CLAUDE.md`, `docs/manual-smoke-tests.md` — documentation surfaces (env-var table, v1.0+ stable-surfaces list, smoke checklist).

## Test plan

- [x] `uv run pytest -q --cov-fail-under=80` — 349 passed (was 335; +14 tests for the new env var, the disabled-timer skip, the cfg → module wiring, and the duration formatter).
- [x] `uv run ruff check && uv run ruff format --check && uv run ty check` all green.
- [ ] Manual smoke test (operator-side): set `RYZIC_AUTOLEAVE_SECONDS=0`, drain a queue, verify the bot stays connected. Restore default afterwards.


riohno:
## Verdict: Approve with nits

Implementation hits the spec cleanly. The code addresses all of issue #62's hard lines (default `300` preserves behaviour; `0` disables; asymmetric scope vs. empty-channel timer documented in both PR body and commit body). 349 tests pass locally, 92.26% coverage, ruff/format/ty all green. **No blockers.** Two small nits below worth addressing before merge; the rest is suggestion-grade.

---

### 1. `_auto_leave_seconds_reset` helper is needless indirection — inline it (nit)

`src/ryzic/lavalink_glue.py:597-609` introduces a top-level `_auto_leave_seconds_reset()` helper that is called from exactly one place (`_reset_state_for_test`) and contains exactly two lines. Worse, the `_set_auto_leave_seconds_for_test(seconds)` function right above it does the same thing parametrically. Two competing knobs for the same concept.

KISS / SRP: collapse to one of:

```python
def _reset_state_for_test() -> None:
    for task in auto_leave_tasks.values():
        if not task.done():
            task.cancel()
    last_play_channel.clear()
    auto_leave_tasks.clear()
    _set_auto_leave_seconds_for_test(DEFAULT_AUTO_LEAVE_SECONDS)
    _voice_ready_events.clear()
```

…which removes `_auto_leave_seconds_reset` entirely. Same behaviour, one less symbol.

### 2. Sentinel handling: `<= 0` is the right choice (no change needed)

The audit asked whether `_auto_leave_seconds <= 0` should be `== 0` since the parser already rejects negatives. **`<= 0` is correct as-is** — defense in depth pairs naturally with the test-only seam (`_set_auto_leave_seconds_for_test` allows negatives without re-validation), and it costs nothing semantically. Worth a one-line comment at line 143/364 noting "negatives can only arrive via the test seam; treat as disabled rather than crashing on `asyncio.sleep(-N)`" if you want to make the intent explicit, but not required.

### 3. Mid-flight reconfig: cancel-then-conditionally-schedule is solid (no change)

`_start_auto_leave` cancels the existing timer **before** the early-return on `_auto_leave_seconds <= 0` (lines 142-144). This is the right shape for the documented "operator restarts with `RYZIC_AUTOLEAVE_SECONDS=0` mid-queue" scenario, and `test_auto_leave_seconds_zero_still_cancels_existing_timer` exercises exactly that invariant. Order is load-bearing; consider keeping the test name explicit about the cancel-first ordering if it ever drifts.

### 4. `_auto_leave(seconds: int)` parameter capture (no change, but worth checking the test rationale)

The audit flagged this as potential over-engineering. In practice it's harmless — `_auto_leave_seconds` is read once at startup, so capturing at task creation vs. at sleep wake-up makes no observable difference today. The defence is that **if** someone later mutates `_auto_leave_seconds` between `create_task` and `await asyncio.sleep`, the captured value preserves the message text against drift. That's worth ~5 LOC.

Minor: the name `seconds` in `_auto_leave(bot, guild_id, seconds)` shadows the variable `seconds` that callers might use locally; consider `idle_seconds` or `window_seconds` for clarity. Pure suggestion.

### 5. `_format_idle_duration` boundaries — coverage gaps (suggestion)

Verified rendering across the natural range:

```
   1s -> '1 second'
  45s -> '45 seconds'
  59s -> '59 seconds'
  60s -> '1 minute'
  61s -> '61 seconds'
  90s -> '90 seconds'
 120s -> '2 minutes'
 300s -> '5 minutes'
3600s -> '60 minutes'
3601s -> '3601 seconds'
7200s -> '120 minutes'
```

`90s -> '90 seconds'` (not "1.5 minutes") and `3600s -> '60 minutes'` (not "1 hour") are both fine — the function deliberately avoids fractional units and hours. `3601s -> '3601 seconds'` is the only mildly ugly edge case, but no operator is realistically going to set `RYZIC_AUTOLEAVE_SECONDS=3601`.

The test (`test_format_idle_duration_renders_minutes_and_seconds`, line 504-509) covers 300, 60, 45, 1 — both singular paths and one each of the plural minute/second branches. Consider adding `_format_idle_duration(120) == "2 minutes"` so the plural-minute branch is covered explicitly (currently only "5 minutes" exercises plural-minutes, and the test file would still pass if the formula `f"{minutes} minute"` regressed to singular for plural). Optional.

### 6. Module-level `_auto_leave_seconds` singleton + reset isolation (no change)

The singleton mirrors the `_ll_client` pattern correctly. `_reset_state_for_test` (autouse fixture in `tests/test_lavalink_bridge.py:135-138`) restores the default on every test, so cross-test pollution is impossible. The custom-value test (`test_auto_leave_seconds_custom_value_passed_into_sleep`) sets `45` then completes the task before the next test runs, and the autouse reset would catch it anyway. Good.

### 7. Asymmetric outcome documented (no change)

Verified:
- PR body has an explicit "Asymmetric outcome (intentional)" section calling out that the empty-channel timer (audit #7, v0.7.0) stays hardcoded.
- Commit body repeats the same: *"NOT cargo-cult symmetry across the two timers."*
- No code touches the empty-channel timer (which doesn't exist in the repo yet — `grep -rn 'empty_channel\|EMPTY_CHANNEL' src/ tests/` returns nothing). Future contributors are explicitly warned not to "while we're here" the symmetry.

Strong execution on this — exactly the documentation discipline the audit asked for.

### 8. README + SEMVER.md updates (no change)

- `README.md:110` has the new row in the env-var table with units, default, and the `0`-disables sentinel.
- `SEMVER.md:25` adds `RYZIC_AUTOLEAVE_SECONDS` to the v1.0+ stable env-var contract list **and** to the units-stability clause (rare to do both correctly on the first pass — nice attention to detail).
- `.env.example` has the entry with a comment explaining the sentinel.
- `docs/manual-smoke-tests.md` adds the optional smoke step for `=0`.
- `CLAUDE.md` updated to reflect the configurable timer.

All five surfaces touched — no missing documentation.

---

### Brittleness flag (suggestion)

`test_register_listeners_installs_auto_leave_seconds_from_cfg` asserts `len(subscriptions) == 4`. This breaks the next time anyone adds a hikari subscription in `register_listeners` (e.g. a `MemberLeaveEvent` listener) for an unrelated reason. The test's actual contract is "the config value reaches the module"; the subscription count is incidental. Consider asserting on the event types subscribed (e.g. `assert hikari.ShardReadyEvent in (e for e, _ in subscriptions)`) rather than the count, OR drop the sanity assertion entirely since the existing voice-bridge tests already cover wiring.

riohno:
## Security review (PR #77)

**Verdict: Approve.** No HIGH or MEDIUM findings. The validation surface is tight, the disable-sentinel is honoured at three layers (`_parse_non_negative_int`, `_start_auto_leave`, `on_queue_end`), and the user-facing message can only contain the operator's own integer — no injection vector for server users. Two LOW notes below for transparency; neither is a blocker.

Threat model items walked through (numbered per the brief):

1. **Input validation** — verified against the seven listed inputs:
   - Empty string → defaults to `300` (matches the project convention used by `RYZIC_GUILD_IDS` / `RYZIC_YOUTUBE_COOKIES_PATH`). Documented in the unit test `test_auto_leave_seconds_empty_defaults_to_300`.
   - `-1` → `ConfigError("RYZIC_AUTOLEAVE_SECONDS must be >= 0, got -1")`. Covered.
   - `0` → accepted and routed through the `<= 0` early-return in `_start_auto_leave`. Covered.
   - `"abc"` → `ConfigError("RYZIC_AUTOLEAVE_SECONDS must be an integer, got 'abc'")`. Covered.
   - `"3.14"` → same `ConfigError` (Python's `int()` rejects float-shaped strings). Covered by the same test.
   - `"99999999999"` → parses fine; Python ints are arbitrary precision; `asyncio.sleep(10**100)` schedules without `OverflowError` (verified empirically). At pathological values (`10**2000` digits of seconds) the rendered Discord message would exceed 2000 chars, but the operator would also be waiting `10**2000` seconds for the message to fire — the bot is effectively never going to disconnect, which is the same observable outcome as `=0`. Not exploitable; not a finding.
   - Whitespace `" 300 "` — `int()` accepts surrounding whitespace, so this works. Reasonable behaviour for an env var.

2. **No secrets in error messages** — `_parse_non_negative_int` only interpolates the env var name and the value passed in (via `{name}` and `{raw!r}` / `{value}`). No adjacent env vars or tokens are leaked. Stack trace via `from exc` adds no sensitive data.

3. **Resource exhaustion (zero-sleep timer storms)** — defended in depth: `_parse_non_negative_int` accepts `0`, `_start_auto_leave` early-returns on `<= 0`, and `on_queue_end` logs the disabled path explicitly. Single-threaded asyncio means there is no race between config load and timer schedule — `register_listeners` runs synchronously at startup before any event loop iteration that could fire `on_queue_end`.

4. **Information disclosure via `_format_idle_duration`** — the function takes an `int` and returns a string built from `{int} {minutes|seconds}`. Since the input is type-constrained to `int` by the dataclass (validated at parse time), the output cannot contain ANSI codes, Discord markdown control chars, or pings. The format string interpolates only into a static prefix/suffix that ryzic itself controls. No injection vector for server users.

5. **Module-level singleton race conditions** — `_auto_leave_seconds` is a module-level `int`. Single-threaded asyncio rules out thread races. `register_listeners` is documented as non-idempotent (consistent with the pre-existing `_ll_client` singleton); a hot reload would reset the value but no security-sensitive state is shared.

6. **Test isolation via `_reset_state_for_test`** — verified: the helper now calls `_auto_leave_seconds_reset()` which sets `_auto_leave_seconds = DEFAULT_AUTO_LEAVE_SECONDS`. The autouse fixture in `tests/test_lavalink_bridge.py:135` calls this before every test, so `pytest-randomly` cannot leak the value across cases. All four tests that mutate the value (`_set_auto_leave_seconds_for_test(0)` / `(45)`) execute inside that fixture's scope.

7. **Singleton pattern security implications** — `_auto_leave_seconds` mirrors `_ll_client` *structurally* (module-level mutable, `global` keyword, test-only setter) but stores no security-sensitive state — just an int. There is no path where the auto-leave singleton's pattern accidentally shares Lavalink client auth state; the two singletons live in independent module-level slots and are never aliased.

### Findings

#### LOW — confidence 7/10 — `_auto_leave_seconds_reset` adds an unnecessary indirection

`_reset_state_for_test` calls `_auto_leave_seconds_reset()` which is a one-line helper that exists only because Python requires the `global` declaration in the function that does the assignment. Inlining the global declaration into `_reset_state_for_test` itself would remove the helper entirely:

```python
def _reset_state_for_test() -> None:
    global _auto_leave_seconds
    for task in auto_leave_tasks.values():
        if not task.done():
            task.cancel()
    last_play_channel.clear()
    auto_leave_tasks.clear()
    _auto_leave_seconds = DEFAULT_AUTO_LEAVE_SECONDS
    _voice_ready_events.clear()
```

This is a `/simplify` concern more than a security one — the current code is correct, just more surface area than needed. Not a blocker.

#### LOW — confidence 6/10 — `_format_idle_duration(0)` would render "0 seconds"

If a future refactor accidentally calls `_format_idle_duration(0)` (currently impossible because the `<= 0` early-return prevents the timer from firing), the output would be `"Idle for 0 seconds — disconnecting."` — semantically nonsense but harmless. The function has no defensive guard against `seconds <= 0`. Currently relied upon by the call-site invariant; future maintainers should be aware. Adding `assert seconds > 0` at the top of `_format_idle_duration` would make the contract explicit. Not a blocker.

### Approval

Approving. Validation is complete and tight, defenses are layered, no secrets leak, and test isolation holds. Ship it.

riohno:
Closing temporarily to retrigger CI — pull_request.synchronize webhooks have been silently missed for the past 3 push attempts on this branch. Will reopen immediately.

riohno:
Reopened to fire fresh pull_request.reopened event for CI.

PR #212, opened 2026-06-23
test: close v0.6.0 behavior-change coverage gaps (#205)

### LINKED ISSUE #205: test coverage gaps on v0.6.0 behavior-change logic (disconnect-marker error paths, refresh_all filter)
## Problem

Several v0.6.0 behavior-change code paths have no effective test coverage: the `refresh_all` per-guild skip filter and stagger are patched out in the only test that touches the loop, the #198 disconnect-marker error paths are unexercised because `FakeBot.update_voice_state` never raises, the #159 ephemeral-default test asserts the function default rather than the lightbulb option default (the layer that affects production), the loop test patches `asyncio.sleep` globally (brittle and process-wide), the i18nice helper's positive `%{var}` interpolation success path is not directly asserted, and `_configure()` is non-idempotent (re-invocation leaves duplicate `load_path` entries — test-state pollution under random ordering).

## Symptoms

1. `tests/test_bot.py` `test_update_controllers_loop_refreshes_active_controllers` patches `ryzic.now_playing.refresh_all` with a counting stub and asserts the loop calls it once. Because `refresh_all` is replaced, the test verifies only loop sleep/cancel plumbing — not the filtering logic (skip when `player is None` / paused / `current is None`), staggering, or per-guild exception isolation. A regression dropping the paused guard (re-rendering paused controllers every 15s, burning REST quota) would pass.
2. `src/ryzic/now_playing.py` `refresh_all` contains the feature's substantive logic (skip paused/idle, stagger, per-guild isolation) but `tests/test_bot.py` patches it out and only asserts the loop calls it once. A bug dropping the paused-skip guard (edit spam, REST quota burns) would pass. The most likely regression class is unguarded.
3. `src/ryzic/commands/leave.py` — #198 adds `_clear_intentional_disconnect` to three error paths but adds no tests. `FakeBot.update_voice_state` is a no-op that never raises, so the try/except branches in `leave.py` and `_auto_leave` are unexercised. `test_guild_leave_clears_per_guild_state` was not updated to assert the marker is cleared. A future refactor dropping a clear call passes silently.
4. `tests/test_queue_command.py` `test_default_response_is_ephemeral` calls `_handle_queue(ctx)` with no `private` kwarg, asserting the function-level default (`private: bool = True`). It does NOT assert `Queue.private` has `default=True`. If a future change flipped the option's `default=` back to `False` while leaving the function signature at `True`, the test passes but production lightbulb passes `private=False` and `/queue` silently reverts to public. The PR description claims "default-flip coverage" but the coverage is of the wrong layer.
5. `tests/test_bot.py` — `patch('asyncio.sleep', side_effect=[None, asyncio.CancelledError()])` replaces the real `asyncio.sleep` process-wide for the test's duration. It works today only because `refresh_all` is mocked (one `asyncio.sleep` call per cycle), but couples the test to the exact internal call count and will break confusingly if the loop adds a second sleep (e.g. a jitter). It also patches a stdlib primitive globally, which can interact badly with pytest-asyncio or background task scheduling.
6. `tests/test_i18n_helper.py` — the suite covers missing key, missing placeholder (failure path), malformed catalog, and locale resolvers, but the smoke key `i18n.smoke` has no placeholder, so the success path of `%{name}` substitution is never directly asserted at the helper level. Positive interpolation is only exercised transitively through `broadcast_t` tests. The helper's core happy path is untested directly; a regression in the substitution logic would only be caught transitively and could be masked.
7. `src/ryzic/i18n/__init__.py` — `_configure()` does `i18n.load_path.append(str(_LOCALES_DIR))` unconditionally at import; `i18n` is a process-global singleton. `tests/test_i18n_helper.py::test_malformed_catalog_falls_back_to_returning_keys` re-invokes `ryzic_i18n._configure()` after monkeypatching `i18n.set` to raise; the `append` runs before the raise, so `load_path` permanently contains a duplicate `_LOCALES_DIR` entry (monkeypatch teardown does not undo it). Harmless for lookups (first match wins, cached) but leaks across tests under pytest-randomly. The test's name/docstring claim to cover "malformed catalog fallback" but its own comment admits it "really proves is the try/except guard around `_configure()` doesn't crash the bot" — it does not exercise an actual malformed catalog path.

## Scope

In scope: missing test coverage for the `refresh_all` per-guild skip filter / stagger / per-guild isolation, the #198 disconnect-marker error paths, the #159 lightbulb-option-layer ephemeral default, the brittle global `asyncio.sleep` patch, the i18nice helper's positive interpolation path, and `_configure()`'s non-idempotent `load_path` append.

Out of scope: the separate catalog-drift-gate coverage gaps (#204), and the duplicate-controller race itself (#201) — this issue covers the test infra, not the behavior.

- FACT: `FakeBot.update_voice_state` in the test harness is a no-op that never raises; exercising the #198 try/except error paths requires a harness change that lets the stub raise on demand.
- FACT: `refresh_all` is patched out in the only loop test, so any direct coverage of its filter/stagger/isolation must call it without patching and assert its per-guild behavior.
- FACT: The lightbulb option `default=` and the handler function signature default are independent layers; only the option `default=` flows to production via the slash-command registration. A regression test for a default flip must assert the option's `default=`, not the function signature's.
- FACT: `_configure()` appends to a process-global `load_path` unconditionally; idempotency would require a membership check or a `load_path.clear()`-and-rebuild, and any re-invocation in tests must not leak duplicates across random ordering.

## Source

Surfaced during v0.6.0 pre-release review 2026-06-23.

riohno:
## Gaps covered

Closes the remaining v0.6.0 behavior-change test-coverage gaps from #205 (the disconnect-marker error-path gaps were folded into #209's marker tests and are intentionally not duplicated here).

1. **#200 `refresh_all` per-guild skip-filter + isolation — NOT unit-tested.** `test_update_controllers_loop_refreshes_active_controllers` only asserted the loop calls `refresh_all` once via a counting stub; the filter (`skip when player is None / paused / current is None`), stagger, and per-guild exception isolation were unexercised. Added three direct tests against `refresh_all` in `tests/test_now_playing.py`:
   - `test_refresh_all_skips_paused_and_idle_players` — seeds PLAYING / PAUSED / IDLE players across three guilds and asserts only the PLAYING guild's controller is edited.
   - `test_refresh_all_skips_guilds_with_no_player` — a controller record with no backing player is skipped.
   - `test_refresh_all_isolates_per_guild_exceptions` — one guild's raising `refresh()` does not abort the loop for the other guild.

2. **#159 `/queue` option-default — test asserted the wrong layer.** `test_default_response_is_ephemeral` only exercised the `_handle_queue` function-signature default (`private: bool = True`); the lightbulb option's `default=True` (the value actually passed at runtime) was unchecked. Added `test_queue_option_private_defaults_to_true` asserting `Queue._command_data.options["private"].default is True` (mirrors the replay loader's `_command_data.options` pattern). The existing function-level test is retained — both layers matter.

3. **#153 i18nice positive `%{var}` interpolation — not directly tested at helper level.** `i18n.smoke` has no placeholder, so the success path of `%{var}` substitution was only exercised transitively. Added `test_positive_placeholder_interpolation_renders_value` in `tests/test_i18n_helper.py` asserting `t("ytdlp.error.generic_with_detail", detail="boom")` renders the interpolated value (uses a real `en_US.json` key with `%{detail}` — no test-only catalog mutation).

4. **#153 `_configure()` non-idempotent + misnamed test.**
   - (a) `src/ryzic/i18n/__init__.py`: guarded the `load_path.append` with a membership check so repeated `_configure()` calls don't accumulate duplicate `_LOCALES_DIR` entries. Behavior-preserving one-line src/ change — the only src/ change in this PR.
   - (b) `tests/test_i18n_helper.py`: renamed `test_malformed_catalog_falls_back_to_returning_keys` → `test_configure_does_not_crash_when_i18n_set_raises` with an honest docstring explaining the rename (the old name claimed a malformed-catalog exercise the test never performed; a true malformed-catalog test would require re-pointing the process-global `load_path` and risks cross-test pollution for marginal extra coverage — the try/except is the actual safety net).

5. **#200 `asyncio.sleep` global-patch brittleness.** Replaced the `side_effect=[None, asyncio.CancelledError()]` list (which coupled the test to the loop's exact internal sleep call count and patched a stdlib primitive process-wide) with a no-wait sleep stand-in plus task-self-cancellation from inside the `refresh_all` stub. The test no longer depends on the number of sleep calls per cycle (a future jitter or second sleep won't break it). The loop calls `asyncio.sleep` via the `asyncio` module namespace, so patching `ryzic.bot.asyncio.sleep` would be identical to patching `asyncio.sleep` — without a src/ change (out of scope) there is no narrower module-local reference to target; the `with`-scoped patch keeps the override local to this test. Rationale documented in the test docstring.

## src/ change

One-line idempotency guard in `src/ryzic/i18n/__init__.py::_configure()`:

```python
if str(_LOCALES_DIR) not in i18n.load_path:
    i18n.load_path.append(str(_LOCALES_DIR))
```

i18nice is a process-global singleton; `load_path` persisted across `_configure()` re-invocations (tests re-run it), accumulating duplicate entries. Behavior-preserving — the first call appends exactly as before; subsequent calls are no-ops. No other src/ changes.

## Closes #205

### DIFF
--- a/src/ryzic/i18n/__init__.py
+++ b/src/ryzic/i18n/__init__.py
@@ -36,7 +36,11 @@ def _on_missing_placeholder(key: str, locale: str, _template: str, name: str) ->
 
 def _configure() -> None:
     try:
-        i18n.load_path.append(str(_LOCALES_DIR))
+        # i18nice is a process-global singleton; ``load_path`` persists
+        # across ``_configure()`` re-invocations (tests re-run it). Guard
+        # the append so repeated calls don't accumulate duplicate entries.
+        if str(_LOCALES_DIR) not in i18n.load_path:
+            i18n.load_path.append(str(_LOCALES_DIR))
         i18n.set("filename_format", "{locale}.{format}")
         i18n.set("file_format", "json")
         i18n.set("fallback", _DEFAULT_LOCALE)
--- a/tests/test_bot.py
+++ b/tests/test_bot.py
@@ -199,21 +199,50 @@ def test_install_cookies_creates_cache_dir_if_missing(tmp_path: Path) -> None:
 
 @pytest.mark.asyncio
 async def test_update_controllers_loop_refreshes_active_controllers() -> None:
-    """The loop calls refresh_all once per cycle and exits on cancellation."""
+    """The loop calls refresh_all once per cycle and exits on cancellation.
+
+    The loop's first action is ``await asyncio.sleep(15)``, so the sleep
+    must be short-circuited to keep the test fast. We replace
+    ``asyncio.sleep`` with a no-wait stand-in that yields once (so a
+    pending task cancellation is delivered at that await rather than
+    spinning the loop synchronously). Cancellation is driven by the
+    ``refresh_all`` stub itself via ``asyncio.current_task().cancel()``
+    after the first call, rather than a ``side_effect`` list of exact
+    length — this decouples the test from the loop's internal sleep call
+    count (a future jitter or second sleep would not break it).
+
+    The loop calls ``asyncio.sleep`` via the ``asyncio`` module namespace
+    (``import asyncio`` + ``asyncio.sleep``), so patching
+    ``ryzic.bot.asyncio.sleep`` would be identical to patching
+    ``asyncio.sleep`` — there's no narrower module-local reference to
+    target without a src/ change, which is out of scope for #205. The
+    ``with``-scoped ``patch`` keeps the override local to this test;
+    pytest-asyncio runs tests sequentially so no sibling coroutine sees
+    the fake.
+    """
     bot_mock = MagicMock(spec=hikari.GatewayBot)
-
-    # refresh_all is patched out, so each cycle is exactly one sleep + one
-    # refresh_all. The second sleep raises CancelledError to break the loop
-    # after one completed iteration.
     call_count = 0
+    real_sleep = asyncio.sleep
 
-    async def counting_refresh_all(bot):
+    async def _no_wait_sleep(_delay: float, *_a: object, **_kw: object) -> None:
+        # Yield once so a pending cancellation lands here instead of the
+        # loop spinning synchronously through the (un-suspending) fake.
+        await real_sleep(0)
+
+    async def counting_refresh_all(_bot: object) -> None:
         nonlocal call_count
         call_count += 1
+        # Cancel the running loop task after the first refresh so the next
+        # await raises CancelledError and the loop breaks. Done here rather
+        # than via a side_effect list so the test doesn't depend on the
+        # exact number of sleep calls per cycle.
+        task = asyncio.current_task()
+        assert task is not None  # we are running inside the loop task
+        task.cancel()
 
     with (
         patch("ryzic.now_playing.refresh_all", side_effect=counting_refresh_all),
-        patch("asyncio.sleep", side_effect=[None, asyncio.CancelledError()]),
+        patch("asyncio.sleep", new=_no_wait_sleep),
     ):
         await bot._update_controllers_loop(bot_mock)
 
--- a/tests/test_i18n_helper.py
+++ b/tests/test_i18n_helper.py
@@ -6,8 +6,8 @@ Covers the four hook surfaces the helper is responsible for:
 - Missing-key fires ``_on_missing_translation`` (logs at ERROR, returns key).
 - Missing-var fires ``_on_missing_placeholder`` (logs at ERROR, returns
   literal ``%{name}``).
-- Malformed-catalog fallback: simulate by re-running ``_configure()`` with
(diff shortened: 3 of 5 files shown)


riohno:
Correction to the PR body: the #198 disconnect-marker error-path tests were described as "folded into #209", but #209's 8 marker tests cover close-event and node-drop clearing — they do **not** make `FakeBot.update_voice_state` raise or assert the `except`-branch clear in `_handle_leave` / `_auto_leave`. That coverage gap is tracked in #217.

The clearing code itself is correct (verified during the v0.6.0 review); this is a test-coverage gap, not a behavior bug, and is not a release blocker.

VeryLongNicknameSuchWow (approved):

Issue #225, opened 2026-06-23
Plan: unify all teardown paths into _teardown_player_session (#218, #224)


riohno:
## Problem

Four distinct teardown paths exist for ending a player session. Only three go through the shared `_teardown_player_session` helper extracted by #209. The fourth (`/leave`) duplicates cleanup steps and misses two. The auto-leave path delegates cleanup to the 4014 close event, which can be lost.

## Scope

Unify all four teardown paths through `_teardown_player_session` so every path releases the same set of resources:

1. **`/leave` (`commands/leave.py:_handle_leave`)** — replace the duplicated `clear_queue_releasing` + `now_playing.teardown` calls with a single `_teardown_player_session` call. Keep `player.stop()` and `mark_intentional_disconnect` + `update_voice_state(None)` as caller-specific work before the helper. This picks up `_cancel_auto_leave` and `_reset_voice_ready` for free, fixing #218.

2. **Auto-leave (`lavalink_glue.py:_auto_leave`)** — call `_teardown_player_session` after `update_voice_state(None)` so cleanup does not depend on the 4014 close arriving. Keep `mark_intentional_disconnect` + `update_voice_state(None)` + the auto-leave broadcast as caller-specific work. This fixes #224.

3. **4014 paths (`on_websocket_closed`)** — no change needed. When the 4014 arrives after `/leave` or `_auto_leave` already called `_teardown_player_session`, the intentional marker is still pending, so `on_websocket_closed` calls `_teardown_player_session` again. This is safe because every step is idempotent (verified):
   - `clear_queue_releasing(player)` — empty queue = no-op
   - `_cancel_auto_leave(guild_id)` — no task = no-op
   - `_reset_voice_ready(guild_id)` — not in dict = no-op
   - `now_playing.teardown(bot, guild_id)` — `_controllers.pop(guild_id, None)` returns immediately if already torn down

4. **`now_playing_buttons.py`** — no change needed. The Stop button dispatches to `_handle_leave`, so fixing `_handle_leave` fixes the button path too.

5. **`_on_guild_leave`** — no change needed. This path already does its own cleanup (no REST teardown, just local state drops) because the bot lost guild permissions. It is intentionally distinct.

## Constraints to surface for planning

- **Double-teardown is safe**: confirmed by reading `now_playing.teardown` (line 328: `_controllers.pop(guild_id, None)` → early return) and every other step in `_teardown_player_session`. All are idempotent.
[...]

PR #226, opened 2026-06-23
fix(teardown): unify /leave and auto-leave through _teardown_player_session

### LINKED ISSUE #218: /leave is a 4th teardown path not folded into _teardown_player_session; does not cancel the auto-leave timer
## Problem
`/leave` runs its own teardown cleanup and is not folded into the shared `_teardown_player_session` helper that #209 extracted for the three lavalink-side teardown paths (4014 intentional, 4014 unintentional, node-disconnected). It also does not call `_cancel_auto_leave`, so a pending auto-leave timer could fire after an explicit `/leave`.

## Symptoms
- Asymmetric teardown: a future change to `_teardown_player_session` (e.g. adding a new cleanup step) will silently not apply to `/leave`, so `/leave` and the auto-leave timer can diverge in what they release (queue pins, now-playing controller, auto-leave timer cancellation).
- A pending auto-leave timer is not cancelled on explicit `/leave`, so it could fire after the user already left.

## Scope
Fold `/leave`'s cleanup into the shared `_teardown_player_session` helper (or coordinate the call sites so they release the same set of resources), and cancel any pending auto-leave timer on explicit `/leave`.

- The existing `/leave` cleanup is functionally adequate for the explicit-leave case — this is a DRY/asymmetry gap + a possible stale-timer fire, not a current behavior bug.
- `/leave` and the auto-leave path may intentionally differ in some respect (e.g. `/leave` is user-initiated, auto-leave is timer-driven) — verify which steps must stay distinct before collapsing.
- `_teardown_player_session` lives in `src/ryzic/lavalink_glue.py`; `/leave` is in `src/ryzic/commands/leave.py`.

## Source
Surfaced during the v0.6.0 final review 2026-06-23. Refs #209.

### LINKED ISSUE #224: auto-leave teardown depends on the 4014 close; a lost 4014 leaks audio-cache pins and orphans the controller
## Problem
`_auto_leave` (`src/ryzic/lavalink_glue.py`) relies on receiving the 4014 websocket close to trigger `_teardown_player_session` (queue-pin release + now-playing teardown). Unlike `/leave`, which cleans up directly and is resilient to a lost close event, a lost 4014 during an auto-leave orphans state (audio-cache pins not released, now-playing controller not torn down).

## Symptoms
Lost websocket close events are rare but possible. An auto-leave that loses its 4014 leaks audio-cache pins and orphans the now-playing controller until a subsequent real disconnect / restart clears it.

## Scope
Make the auto-leave teardown resilient to a lost 4014 — clear on a timeout, or clear via the same direct path `/leave` uses, so the teardown does not depend solely on the 4014 close event arriving.

- The 30s TTL marker GC (#209) bounds the spurious-`voice_lost` consequence, but it is not confirmed to bound the audio-cache pin / controller orphan leak — verify what clears the leaked state today.
- Related to #218 (teardown asymmetry: `/leave` does not call `_cancel_auto_leave`) but distinct — #218 is `/leave`-side; this is the auto-leave path's 4014-dependence.

## Source
Surfaced during the v0.6.0 final review 2026-06-23. Refs #209, #218.

riohno:
## Summary

Unify the two remaining teardown paths (`/leave` and `_auto_leave`) through the shared `_teardown_player_session` helper extracted by #209, so every teardown route releases the same set of resources. `/leave` was duplicating cleanup steps and missing `_cancel_auto_leave` (a pending auto-leave timer could fire after explicit `/leave`). `_auto_leave` relied on the 4014 WebSocket close event to trigger cleanup — if the close was lost, audio-cache pins leaked and the now-playing controller was orphaned.

`/leave` now also takes ownership of teardown *before its first await* — cancelling the auto-leave timer outright and setting an explicit-leave marker that `_auto_leave` re-checks after its `update_voice_state` await. The cancel alone handles a timer whose sleep is still pending during `/leave`'s awaits; the marker is the backstop for the dispatch-boundary edge where the timer's sleep completes in the same event-loop tick as the `/leave` interaction and the loop runs `_auto_leave` first (it self-pops before `/leave` runs, so the dict-keyed cancel cannot reach it). Together they close #218 without relying on the late cancel bundled inside `_teardown_player_session`.

## What's in / what's out

**In:**
- `/leave` calls `_teardown_player_session` instead of duplicating `clear_queue_releasing` + `now_playing.teardown` (picks up `_cancel_auto_leave` and `_reset_voice_ready` for free)
- `_auto_leave` calls `_teardown_player_session` after `update_voice_state(None)`, replacing the bare `_reset_voice_ready` call
- `/leave` cancels the auto-leave timer and calls `begin_explicit_leave` before its first await (in a `try/finally` that clears the marker); `_auto_leave` bows out via that marker instead of broadcasting after an explicit leave (#218)
- 7 new regression tests: teardown helper call verification, auto-leave timer cancellation, idempotency, 4014-after-auto-leave no-op, plus two deterministic #218 race tests — one for a timer that wakes during `/leave`'s first yield (fails without the up-front cancel), one for a timer already past its sleep and self-popped (fails without the explicit-leave guard)

**Out (deferred):**
- Renaming `_teardown_player_session` to public (drop underscore) — cross-module internal calls are already established (`clear_queue_releasing` is also internal)
- Latent double-broadcast if `update_voice_state` raises in `_auto_leave` (pre-existing behavior, not introduced by this diff)
- 4014 paths in `on_websocket_closed` — already call `_teardown_player_session`, no change needed
- `_on_guild_leave` — intentionally distinct (no REST teardown; releases the same set inline), not changed
- `now_playing_buttons.py` — dispatches through `_handle_leave`, fixed transitively

## Test plan

- [x] `uv run pytest -q` — 609 passed, 1 deselected
- [x] `uv run ruff check . && uv run ruff format --check .` — clean
- [x] CI on this PR — green

## Closes / refs

Closes #218, Closes #224. Refs #209, #225.

### DIFF
--- a/src/ryzic/commands/leave.py
+++ b/src/ryzic/commands/leave.py
@@ -1,11 +1,12 @@
 """``/leave`` slash command (M1 §3).
 
-Stops playback, clears the queue, and disconnects from voice. The
-existing ``QueueEndEvent`` auto-leave timer is rendered moot by the
-explicit disconnect; ``_on_voice_state_update`` in ``lavalink_glue``
-clears ``_voice_ready_events[guild_id]`` once the gateway confirms our
-own state went to ``channel_id is None``, so a follow-up ``/play``
-correctly waits on a fresh handshake.
+Stops playback, clears the queue, and disconnects from voice. Cleanup
+goes through ``_teardown_player_session``, the same helper used by the
+auto-leave timer and the 4014 close paths; the guild-leave path
+releases the same set inline (a REST teardown would 403 once the bot
+has lost the guild). The shared set: queued audio-cache pins, the
+now-playing controller, the auto-leave task, the voice-ready handshake
+event.
 """
 
 from __future__ import annotations
@@ -15,7 +16,7 @@ from typing import cast
 import hikari
 import lightbulb
 
-from .. import lavalink_glue, now_playing
+from .. import lavalink_glue
 from ..i18n import locale_for_ephemeral, locale_for_public, t
 from ..voice_check import ensure_same_voice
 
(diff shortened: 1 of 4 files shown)


riohno:
## Hermes Agent Review

**Verdict: Passed** — no security or logic errors found.

Independent reviewer subagent (fresh context, no shared bias with the implementer) reviewed the full diff. Static security scan also clean (no secrets, injection, eval/exec, pickle, or SQL injection patterns).

### Security
- No hardcoded secrets, credentials, or API keys
- No shell injection, SQL injection, or path traversal
- No eval()/exec() with user input
- File IO (audio cache release) uses existing safe patterns — no new file paths constructed from user input

### Logic
- Double-teardown is safe — every step in `_teardown_player_session` is idempotent (verified by test + code reading)
- `/leave` now cancels pending auto-leave timer (regression for #218)
- `_auto_leave` no longer depends on 4014 close event arriving (regression for #224)
- 4014 paths unchanged — already call `_teardown_player_session`

### Non-blocking observations
1. `_teardown_player_session` is "private" (underscore) but called cross-module from `leave.py` — established pattern (`clear_queue_releasing` is also internal), not worth a rename churn
2. Latent double-broadcast if `update_voice_state` raises in `_auto_leave` (pre-existing behavior, not introduced by this diff) — worth a follow-up issue

### Tests
- 607 passed, 1 deselected (5 new regression tests)
- ruff check + format: clean
- ty check: clean
- CI: all green

---
*Reviewed by Hermes Agent (independent subagent + static scan)*

riohno:
## Hermes Agent Security Review

**Verdict: Clean** — no security concerns found.

### Scan coverage
- Static scan: hardcoded secrets, shell injection, eval/exec, pickle deserialization, SQL injection — all clean
- Independent reviewer: no security concerns identified

### Touch points reviewed
- **File IO (audio cache release)**: `clear_queue_releasing` walks `player.queue` and calls `_release_track` → `cache.release(video_id)`. The `video_id` is derived from `Path(track.identifier).stem` where `identifier` is a Lavalink `LocalAudioSourceManager` file path. No user-supplied input reaches this path — identifiers are set by `ytdlp.py` during download. No path traversal risk.
- **WebSocket close events**: `on_websocket_closed` reads `event.code` (int) and `event.reason` (string). The reason is only logged, never executed or passed to a shell. No injection risk.
- **`update_voice_state(guild_id, None)`**: guild_id is an int from lavalink event payloads, not user input.

### No new attack surface
This PR is a refactor — two call-site swaps routing existing cleanup through an existing helper. No new inputs, no new outputs, no new network calls, no new file operations. All data flows are unchanged.

---
*Reviewed by Hermes Agent*

VeryLongNicknameSuchWow (approved):

Issue #215, opened 2026-06-23
pause/resume/seek/skip: copy + error handling in the Lavalink-disconnected-but-stale-current state


riohno:
## Problem
When Lavalink reports a player as disconnected (`is_connected=False`) but still holds a `current` track reference — a transient resync window that occurs during voice region migration or a voice-WebSocket blip — the pause/resume/seek/skip commands have two problems:

1. **Unverified error handling.** Per #196's guard change (confirmed by the #202 investigation), the commands PROCEED in this state — they call `player.set_pause` / `seek` / `skip` on a player Lavalink considers disconnected. Whether lavalink 5.11.0 raises, no-ops, or silently drops the command in this state is untested, and the command paths' handling of a disconnected-player error is unverified. If lavalink raises and a command path doesn't catch it, the user could see an unhandled error; if it no-ops, the user sees a success response ("Paused.") with nothing actually paused.

2. **Misleading copy.** The state is neither "nothing is playing" (there IS a `current` track) nor a healthy playing state. The current proceed path gives no indication the bot is reconnecting; the alternative reject path (the pre-#196 guard) would say "Nothing is playing", which is false here. Neither copy accurately describes what is happening.

## Symptoms
- Repro: a lavalink `DefaultPlayer` with `is_connected=False`, `current=<track>`. Invoke `/pause` (or `/resume` / `/seek` / `/skip`). Current behavior: the command proceeds with a success response; whether the server-side action lands is unverified.
- The state is transient and rare (seconds during resync), but when hit the user gets no feedback that the bot is reconnecting.

## Scope
The guard behavior and user-facing copy for the `is_connected=False, current is not None` state across `pause.py`, `resume.py`, `seek.py`, `skip.py` (and any shared guard helper). Verifying/characterizing the command paths' error handling when commanding a disconnected lavalink player.

## Constraints to surface for planning
- **lavalink 5.11.0 behavior**: does `DefaultPlayer.set_pause` / `set_seek` / `skip` raise, no-op, or send-and-ignore when `is_connected=False`? The fix's safety depends on this — characterize before designing.
- **DRY**: the four command files share the guard pattern (`if player is None or player.current is None: …`). A new state branch should likely live in a shared helper (the codebase already has `voice_check.ensure_same_voice` etc.) rather than be duplicated four times.
[...]

PR #229, opened 2026-06-24
fix(now_playing): disable controller buttons while player is disconnected

### LINKED ISSUE #222: now_playing render path ignores is_connected, contradicting #215 reconnecting semantic
## Problem
#215 made pause/resume/seek/skip reject with "Reconnecting to voice" when the Lavalink player is disconnected (`is_connected=False`) but holds a `current` track. The now_playing render/refresh path (`src/ryzic/now_playing.py`) does not consult `is_connected`, so the now-playing controller can render a track as playable (with active buttons) while the commands reject with "Reconnecting to voice." User-facing inconsistency within the 30s TTL resync window.

## Symptoms
During the transient resync window (`is_connected=False`, `current` held): the controller shows the track as playable + the buttons appear live, but `/pause` / `/resume` / `/seek` / `/skip` reject with "Reconnecting to voice."

## Scope
Reconcile the now_playing render path with #215's reconnecting semantic — render a reconnecting/disabled state on the controller when `is_connected=False`, or document the intentional divergence.

- The window is the 30s TTL blip (rare, transient); low impact.
- #215's `check_player_or_respond` is the command-side guard; the now_playing render/refresh path is separate (`src/ryzic/now_playing.py`).
- The periodic refresh (#172/#200) re-renders the controller on a timer; the render path's `is_connected` handling must compose with it.

## Source
Surfaced during the v0.6.0 final review 2026-06-23. Refs #215, #172, #200.

riohno:
## Summary

`_render_for_player` ignored `is_connected`, so during the #215 resync window (`is_connected=False` with a held track) the now-playing controller rendered live buttons while `/pause` `/resume` `/seek` `/skip` reject with "Reconnecting to voice" — a user-facing inconsistency within the ~30s resync window. The render path now disables all buttons when the player is disconnected, and `refresh_all`'s paused-skip is narrowed to the connected case so the disabled render applies to both playing and paused players.

## What's in / what's out

**In:**
- `_render_for_player` reuses the held-track embed (`build_now_playing_embed`) — not the idle embed, whose "No tracks playing" copy is factually wrong because `current` is still retained per #215 — paired with `_build_idle_components()` (all buttons disabled), with **disconnected → paused → playing** precedence.
- `refresh_all`'s skip predicate narrowed to `player is None or player.current is None or (player.paused and player.is_connected)`, so a disconnected player (playing *or* paused) is re-rendered with disabled buttons; a playing player auto-recovers to enabled on the next cycle once `is_connected` returns True.

**Out (deferred):**
- A paused player that reconnects while still paused is skipped again by `refresh_all`, so its controller stays disabled until the next `/resume` *after reconnection* — a bounded, safer transient than the #222 defect (disabled-when-should-be-enabled vs the original enabled-when-should-be-disabled). No recovery hook is added (the design uses the periodic loop); pinned by a test as an intentional limitation.
- A dedicated "Reconnecting" embed variant (frozen progress bar) — rejected per KISS for a transient window.
- `FakePlayer.is_connected` fidelity vs lavalink's read-only `channel_id`-derived property — non-blocking, no current test trips it.

No new ux builder, no new i18n strings.

## Test plan

- [x] `uv run pytest -q` — 616 passed, 1 deselected
- [x] `uv run ruff check . && uv run ruff format --check .` — pass
- [x] `uv run ty check` — pass
- [ ] (if integration-relevant) `uv run pytest -q -m integration` — N/A (render-path logic, fully unit-tested)
- [ ] CI on this PR — green — `lint-and-test`/`workflow-edits`/`label` green; `docker-build`/`integration`/`Analyze` pending

New tests: `test_refresh_renders_disabled_buttons_when_disconnected`, `test_refresh_all_does_not_skip_disconnected_player`, `test_refresh_all_renders_disabled_for_disconnected_paused` (regression guard — verified to **fail** on the pre-fix predicate and pass after), `test_render_disconnected_takes_precedence_over_paused`, `test_refresh_re_enables_buttons_after_reconnect`, `test_refresh_all_leaves_paused_controller_disabled_until_resume_after_reconnect`, `test_upsert_renders_disabled_buttons_when_disconnected`, plus a `_disabled_in_row` introspection helper. Two rounds of adversarial multi-dimension review (correctness / race / coverage / precedence) drove the predicate-narrowing follow-up.

## Closes / refs

Closes #222.
Refs #215, #172, #200.

### DIFF
--- a/src/ryzic/now_playing.py
+++ b/src/ryzic/now_playing.py
@@ -212,14 +212,29 @@ async def refresh_all(bot: hikari.GatewayBot) -> None:
     """Refresh all active controllers whose progress is actually moving.
 
     Used by the bot's background loop to advance progress bars. Skips
-    paused / idle / disconnected players so we don't issue byte-identical
-    edits every cycle (their state only changes via an event, which
-    already routes through :func:`refresh`). Staggers updates to spread
-    concurrent in-flight REST requests.
+    idle players and paused players that are connected — their render is
+    byte-identical between cycles, so re-editing would waste the Discord
+    per-message edit budget (state only changes via an event, which
+    already routes through :func:`refresh`). A disconnected player (the
+    #215 resync window — ``is_connected=False`` with a held track) is NOT
+    skipped: re-rendering it keeps the controller's buttons disabled via
+    :func:`_render_for_player`, closing the #222 gap where the controller
+    advertised live buttons the command side rejects. A playing player
+    auto-recovers to enabled buttons on the next cycle once
+    ``is_connected`` returns True. A paused player is skipped again once it
+    reconnects (paused+connected), so its controller stays disabled until
+    the next user-initiated ``/resume`` *after reconnection* — a bounded
+    transient, strictly safer than the #222 defect (disabled-when-should-
+    be-enabled beats enabled-when-should-be-disabled). No separate
+    recovery hook is needed. Staggers updates to spread concurrent
+    in-flight REST requests.
     """
     for guild_id in list(_controllers.keys()):
         player = lavalink_glue.get_player(guild_id)
-        if player is None or player.paused or player.current is None:
+        # Skip paused only when connected: a paused controller is static
+        # (byte-identical re-render), but a disconnected+paused player must
+        # still be re-rendered so its buttons flip to disabled (#222).
+        if player is None or player.current is None or (player.paused and player.is_connected):
             continue
         try:
             await refresh(bot, guild_id)
@@ -253,7 +268,25 @@ async def _render_for_player(bot: hikari.GatewayBot, guild_id: int, channel_id: 
         queue_length=len(player.queue),
         locale="en_US",
     )
-    components = _build_components_paused() if player.paused else _build_components()
+    # Disconnected takes precedence over paused: a player in the #215
+    # resync window (``is_connected=False`` but ``current`` held) must not
+    # advertise a clickable Resume even when paused. The command-side
+    # guard (``voice_check.check_player_or_respond``) rejects
+    # /pause /resume /seek /skip with "Reconnecting to voice" in this
+    # state, so the controller reuses the held-track embed (the idle
+    # embed's "No tracks playing" copy is factually wrong — ``current`` is
+    # still held) paired with all-disabled buttons. ``refresh_all`` does
+    # NOT skip a disconnected player (the paused-skip is narrowed to the
+    # connected case), so a playing player auto-recovers to enabled
+    # buttons once ``is_connected`` returns True; a paused player is
+    # skipped again once reconnected, so its controller stays disabled
+    # until the next user-initiated ``/resume`` after reconnection.
+    if not player.is_connected:
+        components = _build_idle_components()
+    elif player.paused:
+        components = _build_components_paused()
+    else:
+        components = _build_components()
     await _post_or_edit(bot, guild_id, channel_id, embed=embed, components=components)
 
 
--- a/tests/test_now_playing.py
+++ b/tests/test_now_playing.py
@@ -225,6 +225,248 @@ async def test_refresh_recovers_from_deleted_message_by_reposting() -> None:
 
 
 # ---------------------------------------------------------------------------
+# #222 — disconnected render path (is_connected=False with current held)
+# ---------------------------------------------------------------------------
+
+
+async def test_refresh_renders_disabled_buttons_when_disconnected() -> None:
+    """#215/#222: a held track with ``is_connected=False`` is the resync window.
+
+    The command-side guard rejects /pause /resume /seek /skip with
+    "Reconnecting to voice"; the controller must not advertise live buttons
+    either. The held-track embed is reused (NOT the idle embed — ``current``
+    is still held), and every button is disabled.
+    """
+    rest = _FakeRest(create_returns=9700)
+    bot = _bot_with_rest(rest)
+    ll = FakeLavalinkClient()
+    install_lavalink_client(ll)
+    player = ll.player_manager.create(guild_id=111)
+    player.is_connected = True
+    player.current = make_track_with_info(title="Held Track")
+    lavalink_glue.last_play_channel[111] = 555
+    await now_playing.upsert_for_track_start(bot, 111)
+    rest.edit_calls.clear()
+
+    # Lavalink reports disconnected while still holding the track.
+    player.is_connected = False
+    await now_playing.refresh(bot, 111)
+
+    assert len(rest.edit_calls) == 1
+    edit = rest.edit_calls[0]
+    embed: hikari.Embed = edit["embed"]
+    # Held-track embed, not the idle embed (whose "No tracks playing" copy
+    # would be factually wrong — current is retained per #215).
+    assert embed.title == t("ux.np.title.playing", locale="en_US")
+    [row] = edit["components"]
+    assert _disabled_in_row(row) == [True, True, True]
+
+
+async def test_refresh_all_does_not_skip_disconnected_player() -> None:
+    """#222: ``refresh_all`` deliberately re-renders disconnected players.
+
+    The skip predicate does not skip a disconnected player (the paused
+    term is narrowed to the connected case): re-rendering keeps the
+    disabled-buttons visual during the resync window and is what
+    auto-recovers to enabled buttons once ``is_connected`` returns True.
+    Pins the decision so a future "optimization" adding ``is_connected``
+    to the predicate doesn't silently drop the visual.
+    """
+    rest = _FakeRest(create_returns=9800)
+    bot = _bot_with_rest(rest)
+    ll = FakeLavalinkClient()
+    install_lavalink_client(ll)
+
+    # Guild 111: PLAYING (control). Guild 222: disconnected but current held.
+    for guild_id, connected in [(111, True), (222, False)]:
+        player = ll.player_manager.create(guild_id=guild_id)
+        player.is_connected = connected
+        player.current = make_track_with_info()
+        lavalink_glue.last_play_channel[guild_id] = guild_id * 10
+        await now_playing.upsert_for_track_start(bot, guild_id)
+    rest.edit_calls.clear()
+
+    await now_playing.refresh_all(bot)
+
+    edited_channels = {call["channel_id"] for call in rest.edit_calls}
+    # Both re-rendered: the connected one advances progress, the
+    # disconnected one re-renders disabled buttons.
+    assert edited_channels == {111 * 10, 222 * 10}
+    disconnected_edit = next(c for c in rest.edit_calls if c["channel_id"] == 222 * 10)
+    [row] = disconnected_edit["components"]
+    assert _disabled_in_row(row) == [True, True, True]
+
+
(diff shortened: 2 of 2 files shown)


riohno:
needs-changes — the precedence ordering is correct, but the PR's headline auto-recovery only works for disconnected+**playing** players; disconnected+**paused** players are still skipped by `refresh_all`, so #222 persists for that sub-case and the new test gives false coverage.

## Confirmed findings

**#222 is not fixed for disconnected+PAUSED — the new disconnected>paused branch is dead in production** `medium` `precedence` `correctness` `race`
`refresh_all`'s skip predicate at `src/ryzic/now_playing.py:227` (`if player is None or player.paused or player.current is None: continue`) is unchanged by this PR and short-circuits on `player.paused` *before* any `is_connected` consideration. A player paused while connected that then enters the #215 resync window (`is_connected=False`, `current` held, `paused=True`, no 4014 — the non-4014 voice-close path at `lavalink_glue.py:568-576` does not tear down) is skipped every 15s tick, so its controller keeps the PAUSED row with an **enabled Resume button** — the exact #222 visual defect. Every other re-render path is blocked for this sub-case: `/pause` and `/resume` call `check_player_or_respond` (`voice_check.py:100-105`) which rejects when `not is_connected`; `TrackStart`/`QueueEnd` don't fire while paused+disconnected. So the new disconnected branch at `now_playing.py:271-272` is only reachable via a synthetic direct `refresh()` call.
- Evidence: `src/ryzic/now_playing.py:227` (skip predicate), `:271-272` (dead branch for paused), `src/ryzic/commands/pause.py:36-49` & `resume.py:36-49` (post-guard refresh), `src/ryzic/voice_check.py:103-105` (reject), `src/ryzic/lavalink_glue.py:427,465,550` (no refresh on `is_connected` flip), `src/ryzic/bot.py:115-116` (15s loop).
- Suggested action: narrow the paused-skip to the connected case: `if player is None or player.current is None or (player.paused and player.is_connected): continue`. Add a regression test: disconnected+paused guild with an existing controller → `refresh_all` → assert the edit lands with all buttons disabled (currently fails; passes after the fix). Note the 4014 teardown path (`lavalink_glue.py:594-596`) bounds the window in practice, and clicking the stale enabled Resume fails gracefully via the "Reconnecting to voice" ephemeral — so this is a bounded transient visual inconsistency, not a crash/state bug.

**The new precedence test masks the gap — it exercises only the synthetic `refresh()` path** `medium` `precedence` `coverage`
`test_render_disconnected_takes_precedence_over_paused` (`tests/test_now_playing.py:299-324`) calls `now_playing.refresh(bot, 111)` directly at `:319`, bypassing the command guard and `refresh_all`. It proves `_render_for_player`'s precedence is correct *in isolation* but never proves production routes a disconnected+paused player through it. The only `refresh_all` disconnected test (`:265-296`) seeds its disconnected guild with `paused=False` (FakePlayer default), so it also misses the paused subset. Together they give false coverage for the paused sub-case.
- Suggested action: add the `refresh_all`-on-paused+disconnected test described above; it would fail today and expose the gap.

**`refresh_all` docstring overclaims auto-recovery for the paused sub-case** `low` `precedence`
The updated docstring (`now_playing.py:217-223`) and inline comment (`:268-270`) state disconnected players are "deliberately NOT skipped" and "auto-recovers to enabled buttons once `is_connected` returns True — no separate recovery hook is needed." That holds only for disconnected+**playing**. For disconnected+**paused** the `player.paused` term skips every cycle, and even after `is_connected` returns True the player is still paused so `refresh_all` keeps skipping it. The comment mis-describes the predicate a reader is staring at.
- Suggested action: either apply the predicate fix above and keep the claim, or narrow the text to scope the auto-recovery claim to the playing case and call out the paused gap (reference #222).

**No test pins the auto-recovery transition (is_connected False→True re-enables buttons)** `low` `race` `coverage`
The PR's central load-bearing justification for re-rendering disconnected players in `refresh_all` is that buttons "auto-recover to enabled once `is_connected` returns True." No test ever flips `is_connected` back to True. Every `is_connected=False` (`tests/test_now_playing.py:252`, `:318`) is a terminal state. A future regression making the disabled state sticky (cached components, or the "optimization" the docstring warns against) would pass the suite while stranding users with unclickable buttons after reconnect.
- Suggested action: `test_refresh_re_enables_buttons_after_reconnect` — seed connected, flip `is_connected=False` and refresh (assert all-disabled, title `'Playing'`), flip back to `True` and refresh again (assert `_disabled_in_row == [False,False,False]`).

**Disconnected+paused embed title/body is not pinned** `low` `coverage`
The precedence test asserts only `_disabled_in_row`, never `edit['embed'].title`. For disconnected+paused, `build_now_playing_embed` gets `paused=True` so the title is `'Paused'` (`ux.py:529-530`) while buttons are all-disabled idle components — an intentional but unpinned combination. The sibling disconnected+playing test (`:232`) *does* pin the embed title to `'Playing'`, making this gap real. A future "Reconnecting" title variant would change visible copy with no test catching it.
- Suggested action: in the paused precedence test, also assert `edit['embed'].title == t('ux.np.title.paused', locale='en_US')`.

**No test for `upsert_for_track_start` while disconnected** `low` `coverage`
`upsert_for_track_start` calls `_render_for_player` directly (`now_playing.py:177-192`), so the disconnected branch is reachable on the create path, not just refresh. The only upsert-with-held-track test sets `is_connected=True` (`tests/test_now_playing.py:88-105`). Unlikely in production (TrackStart normally implies a voice connection) but the branch is shared and the test is cheap.
- Suggested action: `test_upsert_renders_disabled_buttons_when_disconnected` — `is_connected=False` + held track + `last_play_channel`, call `upsert_for_track_start`, assert all-disabled components and `'Playing'` title (held-track embed, not idle).

**FakePlayer fidelity gap vs lavalink's read-only `channel_id`-derived `is_connected`** `low` `coverage`
Real `lavalink.DefaultPlayer.is_connected` is a read-only `@property` returning `channel_id is not None` (`.venv/.../lavalink/player.py:129-132`), and `position` returns 0 when `not is_playing` (`:134-146`). FakePlayer models `is_connected` as a settable plain bool with no `channel_id` (`tests/_command_helpers.py:152`) and `position` as a plain int with no `is_playing` guard (`:153`). No current test trips it, but a future test could set `position=5000` with `is_connected=False`, pass 5000 to the embed on the fake while production reports 0.
- Suggested action (non-blocking): make `is_connected` a read-only property derived from `channel_id`, and gate `position` on `is_playing`. Not blocking — no current test trips it.

**Disconnected embed body shows "Playing" with a frozen 0:00 progress bar (pre-existing, now owned by the PR)** `low` `correctness`
On a real disconnected player `position` returns 0 (`is_playing == is_connected and current is not None`), so `build_now_playing_embed` renders title `'Playing'` with `'0:00 / <duration>'` and disabled buttons. Not a regression (the pre-PR predicate never skipped disconnected players either), but the PR's updated `refresh_all` docstring now explicitly owns this rendered body as "the correct auto-recovery surface," so the frozen 0:00 under a "Playing" title is mildly misleading for the resync window. Acceptable per KISS (no reconnecting embed variant exists, transient window).
- Suggested action: none required for this PR. If desired later, add a `'Reconnecting'` title variant keyed on a disconnected flag rather than reusing paused/playing.

## What was checked and looked fine

- **Correctness:** The core change is sound — placing `not player.is_connected` ahead of the paused/playing branch in `_render_for_player` correctly disables all buttons during the resync window, and reusing the held-track embed (rather than the idle embed) is correct since `current` is still held. The explicit `refresh()` path is correct for both disconnected+playing and disconnected+paused. The one functional gap is the paused subset via `refresh_all` (above).
- **Race / concurrency:** Single-event-loop asyncio — "races" reduce to interleavings at `await` boundaries. Verified: no suspension between the two `_controllers.get` reads in `refresh` → `_post_or_edit` (`now_playing.py:204` and `:308`) before the REST await, so teardown cannot force a re-register mid-render — it can only interleave at the REST await (last-write-wins on the same message, no re-render loop). Discord's 5-edits/5s cap is unchanged: the skip predicate is byte-identical pre/post PR and 15s ≫ 5s. Cross-tick recovery clobber is unreachable (ticks are strictly sequential, each edit awaited before the 0.1s stagger and next 15s sleep — monotonic). The `is_connected` TOCTOU is benign and the PR flips it to the safer direction (stale-disabled-on-reconnect self-heals in ≤15s, vs the pre-PR stale-enabled-on-disconnect that was the #222 bug itself).
- **Coverage:** The three new tests solidly pin the core change (disconnected renders all-disabled, `refresh_all` does not skip disconnected guilds, disconnected>paused precedence for components), the suite passes (25/25 in `test_now_playing.py`), and `_disabled_in_row` correctly reads the hikari build payload. No pre-existing test encoded the old buggy behavior (main has no disconnected-state tests), so the PR strictly adds coverage. The gaps are the recovery transition and paused embed title (above).
- **Precedence:** The ordering in isolation — idle > disconnected > paused > playing — is correct and mirrors the command-side guard ordering (`voice_check.py:100-105`, no-current before disconnected). Disabled-over-paused is right (`/resume` rejects with "Reconnecting to voice," so an enabled Resume would advertise a non-functional action). Disabled-over-playing is right (`/pause`, `/skip`, `/leave` all reject while disconnected). The precedence is sound as written; the wiring around it (the `refresh_all` skip predicate) is what leaves the paused sub-case uncovered.

riohno:
**Follow-up update** (amended into the single commit): the first ultracode review found the disconnected+**paused** sub-case was not actually fixed — `refresh_all`'s predicate short-circuited on `player.paused` before `is_connected`, so a paused player in the resync window kept an enabled Resume button (the exact #222 defect), and the precedence test gave false coverage by calling `refresh()` directly.

Changes in this amend:
- **Narrowed the `refresh_all` paused-skip to the connected case**: `player is None or player.current is None or (player.paused and player.is_connected)`. Now a disconnected player — playing *or* paused — is re-rendered with disabled buttons.
- **Regression guard**: `test_refresh_all_renders_disabled_for_disconnected_paused` — verified to **fail** on the pre-fix predicate and pass after.
- **Auto-recovery test**: `test_refresh_re_enables_buttons_after_reconnect` pins the playing-case `is_connected` False→True transition through `refresh_all`.
- **Honest docstring**: documents the bounded residual — a paused player's controller stays disabled until the next `/resume` *after reconnection* (inverse polarity of #222, strictly safer, bounded by 4014 teardown). No "byte-identical" overclaim for this window.
- **Pin test** for that residual: `test_refresh_all_leaves_paused_controller_disabled_until_resume_after_reconnect`, so it's an intentional documented limitation, not an unnoticed bug.
- Pinned the disconnected+paused **embed title** (`Paused`) and added an `upsert_for_track_start`-while-disconnected test.

A second adversarial review of the follow-up confirmed **#222 is closed** (no remaining reachable state where the controller advertises an enabled button the command side rejects), no blocker/high findings. The only residual is the bounded paused-reconnect transient above, deliberately deferred (a recovery hook is out of scope for this PR — the design uses the periodic loop).

Checks: `ruff check` ✓, `ruff format --check` ✓, `ty check` ✓, `pytest -q` → **616 passed, 1 deselected**.

VeryLongNicknameSuchWow (approved):

